PR #2, opened 2019-06-11
Bugfix: RCTTVView patches

hufkens:
## Summary

In version 0.59 somehow the `[(RCTRootView *)rootview setReactPreferredFocusedView:self];` line was removed. This caused broke the behaviour of setting the focus to a view.

We are working on an app, and the delay caused all sorts of weird behaviour. We patched this in our version and it's working. I made a PR to the core repo at the time, but nobody knew why the delay was there. And in the end it didn't get merged.

## Changelog

* Fixed broken functionality when you try to set the `hasTVPreferredFocus` through `nativeProps` or `updateView`
* Removed delay in `setHasTVPreferredFocus`

## Test Plan

We have this code running in our app. And for us it's working out.

### DIFF
--- a/React/Views/RCTTVView.m
+++ b/React/Views/RCTTVView.m
@@ -228,7 +228,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:unused)
 {
   _hasTVPreferredFocus = hasTVPreferredFocus;
   if (hasTVPreferredFocus) {
-    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1.0 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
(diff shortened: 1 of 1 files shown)


gcesarmza:
@dlowder-salesforce  This fix was merged into master but it's not included in the 0.59.8-0 release. Are there any chances that it gets merged and a new version is released?

dlowder-salesforce:
Yes... I’ll get this tested and merged into the next release (0.60.4)

gcesarmza:
If you could make a new 0.59.8 release I'd really appreciate it! 🙏

dlowder-salesforce:
Sure!

dlowder-salesforce:
@gcesarmza I just released https://github.com/react-native-community/react-native-tvos/releases/tag/v0.59.8-1 with this fix.

PR #29, opened 2019-11-19
Add information in README.md about common pitfall

haakonnessjoen:
## Summary

Both me and a colleague seperately tried to use react-native-tvos, and both went into the same pitfall of installing react-native globally. And it seems to be a easy pitfall to fall into, as found out after a bit of googling.

## Changelog

Internal Added - Updated README.md to inform about a common pitfall.

## Test Plan

No code changes, only documentation.


dlowder-salesforce:
@haakonnessjoen thanks very much for this -- it's merged and will be in the next release 👍

PR #61, opened 2020-04-04
Refactor RCTTVView's event sender and animation to support subclassing

keshavkaul:
## Summary

**The motivation behind the refactor of focus, blur, & select events**
Currently, select, focus, blur events for a view are broadcasted via TVRemoteHandler Module, which I have experienced to be a major source of frame drops in our app when the number of touchable elements on the screen increases. I have instead subclassed RCTTVView in our app which uses RCTDirectEventBlock to send events to the view.

**The motivation behind the refactors of parallax animation.**
In our app, there are scenarios where the blur event of a touchable element is not called to remove the parallax effects but instead, another focus event is called thereby adding another parallax effect in addition to the already existing one. thereby increasing the size of the touchable element from normal.
To fix this, in our app, we subclassed RCTTVView and added a flag that checks if the parallax effect has already been added therefore need not add it again. 

## Changelog

[tvOS] [Added] - Refactor RCTTVView's event sender and animation to support subclassing

## Test Plan
NA

### DIFF
--- a/React/Views/RCTTVView.h
+++ b/React/Views/RCTTVView.h
@@ -28,4 +28,29 @@
  */
 @property (nonatomic, assign) BOOL hasTVPreferredFocus;
 
+/**
+ * Send Focus Notifications to listeners
+ */
+- (void)sendFocusNotification:(UIFocusUpdateContext *)context;
+
+/**
+ * Send Blur Notifications to listeners
+ */
+- (void)sendBlurNotification:(UIFocusUpdateContext *)context;
+
+/**
+ * Send Select Notification to listeners
+ */
+- (void)sendSelectNotification:(UIGestureRecognizer *)recognizer;
+
+/**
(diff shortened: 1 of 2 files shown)


keshavkaul:
@dlowder-salesforce Hey would be great if you could review this PR about refactoring RCTTVView's event sending and parallax animations to support subclassing of the class. Thanks 👍

dlowder-salesforce:
This is a nice refactor.  Could you add a commit that does what your subclass does -- check to see if the parallax has been added already and not apply it twice?

keshavkaul:
@dlowder-salesforce Thanks! Just pushed a fix for parallax effects being added more than once.
What do you think about using RCTDirectEventBlock for sending focus, blur, and select events to views?

dlowder-salesforce:
@keshavkaul That seems like a good thing to try....

PR #65, opened 2020-04-27
TVFocusGuideView improvements for 0.62.2

dlowder-salesforce:
## Summary

- In 0.62.2, the TVFocusGuideExample in RNTester is not working correctly.  Debugging showed that the destinations prop of the focus guide was being updated too late.  This is fixed by native changes that implement a custom setter for the prop in native.
- The native implementation requires a wrapper view for the focus guide to use in setting its layout.  This PR includes changes to TVFocusGuideView.js to add a wrapper view, so app developers are no longer required to do this.  It also allows the focus guide view to wrap child components, so that the guide can direct navigation from outside to a particular focusable child component.

## Test Plan

Tested with the RNTester TVFocusGuideExample

### DIFF
--- a/Libraries/Components/AppleTV/TVFocusGuideView.js
+++ b/Libraries/Components/AppleTV/TVFocusGuideView.js
@@ -41,10 +41,15 @@ class TVFocusGuideView extends React.Component<FocusGuideProps> {
   render() {
     if (Platform.isTVOS) {
       return (
-        <RNFocusGuide
-          ref={ref => (this._focusGuideRef = ref)}
-          destinationTags={this._destinationTags}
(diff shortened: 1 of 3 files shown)


dlowder-salesforce:
@hufkens are you ok with these changes?

keshavkaul:
@dlowder-salesforce We can avoid adding a wrapper view if just refactor `RNFocusGuide`'s native View implementation to add the focus guide to itself.

One suggestion I have is to expose the `isEnabled` property of UIFocusGuide to support the use case where destinations don't change, but the focus guide is disabled or enabled programmatically.

PR #66, opened 2020-05-04
Fix/android tv overflow and rtl

vdmtrv:
## Summary

Android TV-specific fixes for RTL support and allowing views to overflow beyond the bounds of `ReactHorizontalScrollContainerView`.

## Changelog

[Android TV] [FIXED] - Allowing overflow on ReactHorizontalScrollView
[Android TV] [FIXED] - RTL support for horizontal lists

### Overflow - related to https://github.com/facebook/react-native/pull/22693
* Create a horizontal `<FlatList/>` with styled items having `scale: 2` and a border around each item
* Item border **should not** be cut-off once app is launched

### RTL support
* Create a horizontal `<FlatList/>` that contains a long list of square views
* Use the Android TV developer option to force RTL for all apps
* Open test app and attempt to navigate left/right
* Horizontal list should scroll in desired direction correctly rather than pushing content off-screen

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewProps.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewProps.java
@@ -197,6 +197,7 @@ public class ViewProps {
               FLEX_SHRINK,
               FLEX_WRAP,
               JUSTIFY_CONTENT,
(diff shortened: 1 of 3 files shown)


dlowder-salesforce:
These changes appear to affect both Android phones and TVs.....

PR #101, opened 2020-09-09
feat: nextFocus for iOS

Krisztiaan:
## Summary

I've found android's focus guidance api more powerful, and since it's already there, and possible to implement for iOS (mostly), why not give more options to devs?

## Changelog

[tvOS] [FEAT] - Add directional next focus targetability

## Test Plan

- create 3 views below each other
- target the next one on `nextFocusRight`
- observe the one below getting focus

### DIFF
--- a/React/Views/RCTTVView.m
+++ b/React/Views/RCTTVView.m
@@ -296,33 +296,26 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder : unused)
   object:@{@"eventType":@"blur",@"tag":self.reactTag}];
 }
 
-- (void)setNextFocusUp:(NSNumber *)nextFocusUp {
-  if (nextFocusUp == nil) return;
+- (RCTTVView *)getViewById:(NSNumber *)viewId {
+  if (viewId == nil) return nil;
   NSDictionary<NSNumber *, UIView *> *views = [_bridge.uiManager valueForKey:@"viewRegistry"];
-  RCTTVView *destination = (RCTTVView*)views[nextFocusUp];
-    self->_nextFocusUp = destination;
+  return (RCTTVView*)views[viewId];
+}
+
+- (void)setNextFocusUp:(NSNumber *)nextFocusUp {
+  self->_nextFocusUp = [self getViewById: nextFocusUp];
 }
 
 - (void)setNextFocusDown:(NSNumber *)nextFocusDown {
-  if (nextFocusDown == nil) return;
-  NSDictionary<NSNumber *, UIView *> *views = [_bridge.uiManager valueForKey:@"viewRegistry"];
-  RCTTVView *destination = (RCTTVView*)views[nextFocusDown];
-    self->_nextFocusDown = destination;
+    self->_nextFocusDown = [self getViewById: nextFocusDown];
 }
 
 - (void)setNextFocusLeft:(NSNumber *)nextFocusLeft {
-  if (nextFocusLeft == nil) return;
-  NSDictionary<NSNumber *, UIView *> *views = [_bridge.uiManager valueForKey:@"viewRegistry"];
-  RCTTVView *destination = (RCTTVView*)views[nextFocusLeft];
-    self->_nextFocusLeft = destination;
+    self->_nextFocusLeft = [self getViewById: nextFocusLeft];
 }
 
 - (void)setNextFocusRight:(NSNumber *)nextFocusRight {
-    if (nextFocusRight == nil) return;
-  NSDictionary<NSNumber *, UIView *> *views = [_bridge.uiManager valueForKey:@"viewRegistry"];
-  RCTTVView *destination = (RCTTVView*)views[nextFocusRight];
-
-    self->_nextFocusRight = destination;
+    self->_nextFocusRight = [self getViewById: nextFocusRight];
 }
 
 - (void)setPreferredFocus:(BOOL)hasTVPreferredFocus
--- a/React/Views/RCTTVView.m
+++ b/React/Views/RCTTVView.m
@@ -17,15 +17,17 @@
 #import "RCTUtils.h"
 #import "RCTView.h"
 #import "UIView+React.h"
+#import <React/RCTUIManager.h>
 
 @implementation RCTTVView {
+  __weak RCTBridge *_bridge;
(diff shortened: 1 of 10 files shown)


React/Views/RCTTVView.m
```diff
@@ -266,6 +296,55 @@ - (void)sendBlurNotification:(__unused UIFocusUpdateContext *)context
   object:@{@"eventType":@"blur",@"tag":self.reactTag}];
 }
 
+- (void)setNextFocusUp:(NSNumber *)nextFocusUp {
+  if (nextFocusUp == nil) return;
+  NSDictionary<NSNumber *, UIView *> *views = [_bridge.uiManager valueForKey:@"viewRegistry"];
+  RCTTVView *destination = (RCTTVView*)views[nextFocusUp];
```
dlowder-salesforce, line 302: These two lines are the same in all four setter methods, maybe they should be a single helper method that takes an NSNumber and returns the corresponding view in the registry.

dlowder-salesforce (changes requested):
See inline comments.  Also, it would great if you could extend `RNTester/js/examples/TVFocusGudieViewExample.js` to test the new props.

React/Views/RCTTVView.m
```diff
@@ -266,6 +296,55 @@ - (void)sendBlurNotification:(__unused UIFocusUpdateContext *)context
   object:@{@"eventType":@"blur",@"tag":self.reactTag}];
 }
 
+- (void)setNextFocusUp:(NSNumber *)nextFocusUp {
+  if (nextFocusUp == nil) return;
```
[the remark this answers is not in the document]
Krisztiaan: Good catch

React/Views/RCTTVView.m
Krisztiaan: I wonder if there is already a simple helper for this in RN somewhere?

Krisztiaan:
Converted to draft until I got an example in the sample app.

Krisztiaan:
@dlowder-salesforce I'm wondering if this PR should instead be made against the main `react-native` repo?

olafurnsig:
Isn't nextFocus primarily for TV devices, and since RN kinda supports AndroidTV, but dropped support for tvOS, nextFocus in the main repo is android-only?

Krisztiaan:
While the comments may leave us to believe that, the android documentation lists these props under ["Support keyboard navigation"](https://developer.android.com/training/keyboard-input/navigation). With iOS / iPad already being able to handle keyboard, as well as tvOS, according to their docs, the [necessary api is available everywhere](https://developer.apple.com/documentation/uikit/uifocusenvironment/1648972-preferredfocusenvironments)

dlowder-salesforce:
@Krisztiaan we could do this change in the main repo, but we should refactor it quite a bit first.  If iOS really has full support for all `UIFocusEnvironment` APIs, then we would probably want to take all the focus code and put it in `RCTView` instead of `RCTTVView`, and possibly make other changes as well.  My preference would be to get this change merged here first, and take on the refactor separately before attempting to submit to the core repo.

dlowder-salesforce:
Also @Krisztiaan -- your Typescript extension (#102 ) is already merged.  Are the new `nextFocus` APIs already in `@types/react-native`?  If not, they should be added to `index.d.ts`.

Krisztiaan:
> Also @Krisztiaan -- your Typescript extension (#102 ) is already merged. Are the new `nextFocus` APIs already in `@types/react-native`? If not, they should be added to `index.d.ts`.

Thanks, I was waiting for the other one to get merged first so I could rebase and apply the necessary changes I already have locally.

> My preference would be to get this change merged here first, and take on the refactor separately before attempting to submit to the core repo.

Sounds good, I'll have the example +types some time next week, then think about a good way to approach this.

olafurnsig:
@Krisztiaan I'm very exited for this, any updates? 😄

Krisztiaan:
Wow, I did leave you all hanging! I did not have the capacity the past few weeks, moving to a different project, however I'll try to wrap it up asap, it is not forgotten!

Krisztiaan:
@olafurnsig please give it a try if possible. :)

dlowder-salesforce:
@Krisztiaan thanks very much for this, I'll start reviewing as soon as I'm able

olafurnsig:
I'm a bit lacking in time but I did a quick test with my current project.

Can't be 110% sure since my project is in ultra development mode, and I just tried this on it, it seemed like `tvParallaxProperties` magnification property was just plain ignored on TouchableHiglights, that is, no magnification occured.

Also we're missing the typings for the new properties, so I didn't go much further than that sorry. I'll try to test it again later this week by using the RNtester

acerbetti:
Just testing with my app and it's working very well. 

Thanks for doing it, can't wait to see it merged in

Krisztiaan:
@olafurnsig I've fixed the typings after doing this PR, but I'll add them before merging. Flow types should be up to date though.

I'll have to take a look at the parallax property, thanks for calling it out.

dlowder-salesforce:
@Krisztiaan @olafurnsig I just did some testing with the new RNTester example, and the new focus APIs seem to be working fine 👍 .

Also, magnification seems to be working fine with `tvParallaxProperties`, at least for me.  If that's not the case for you, please send an example and I'll investigate.

dlowder-salesforce (approved):

Krisztiaan:
Same result from me, @olafurnsig an example would be superb.

I'll add the types for this, and if there are no objections, we're good to go :)

dlowder-salesforce:
Thanks @Krisztiaan ... it looks like you added the types, are we ok to merge?

nriccar:
please merge!

douglowder:
@nriccar this has already been merged in https://github.com/react-native-tvos/react-native-tvos/commit/abb6ee1d86b84b155c787ac45aaebdf6b4ab7245

nriccar:
Thank you for the quick response. 

What I don't - I did investigate and read, sorry if its pretty obvious - is what the number stands for. Naturally I tried to pass a ref to the prop nextFocusDown - on every direction.

Am I missing something?

Thanks

Krisztiaan:
Check out the example. It's not a random number, it should be a reference to another view.

Issue #41, opened 2020-01-14
Typescript support


keshavkaul:
I'm using Typescript in my project and was wondering if Typescript support is planned in the near future? I would like to help in this aspect. 

A good first step would be to document all the different TV-related API definitions and then work on adding Typescript support for each API. Thoughts?

dlowder-salesforce:
I just had a look at https://facebook.github.io/react-native/docs/typescript for their recommendations on how to add Typescript support.

It looks like the right place to add definitions for TV-related React Native APIs would be here: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-native

This is the source for the @types/react-native NPM package, which is what FB recommends to use.

Krisztiaan:
If you're using `patch-package`, you can grab [my patch for `@types/react-native-tvos`](https://gist.github.com/Krisztiaan/11856c0faffc70f00766b7f791375a75). I'll look into how it'd be best to add this to `DefinitelyTyped`

olafurnsig:
Upvotes for typescript support!

dlowder-salesforce:
@Krisztiaan thanks VERY much for your patch.  Sorry for the delay in responding here.  If there is anything I can do to help, please let me know.

Krisztiaan:
No, thanks, I'm getting my feet wet with `DefinitelyTyped` and seeing how this'd be best implemented. Probably similar as this repo.

Krisztiaan:
Will update for 63 once we ourselves have done the upgrade too :)

Krisztiaan:
@dlowder-salesforce I'm wondering if an `index.d.ts` file like [this one](https://gist.github.com/Krisztiaan/3d24c7a3c3d4dd8d206c12903adc61d7) would make sense in this repo, so we'd have a lot more flexibility, and a ready to use solution.

dlowder-salesforce:
@Krisztiaan thanks for the PR, looks great!  I'll try making an example TS project that includes most of the TV types.

PR #102, opened 2020-09-09
feat: add Typescript extension

Krisztiaan:
## Summary

Extending `@types/react-native` with `react-native-tvos` specific additions using the already existing flow typings and some research to narrow it down where possible.

## Changelog

[Typescript] [FEAT] - Added Typescript typings

## Test Plan

- try importing `TVEventHandler` in a typescript project

Closes: #41

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,88 @@
+import '@types/react-native';
+
+declare module 'react-native' {
+  import React from 'react';
+
+  import { ViewProps, ScrollViewProps } from '@types/react-native';
+
+  export const useTVEventHandler: (handleEvent: (event: HWKeyEvent) => void) => void;
+
+  export const TVMenuControl: {
+    enableTVMenuKey(): void;
+    disableTVMenuKey(): void;
+  };
+
+  interface HWFocusEvent {
+    eventType: 'blur' | 'focus';
+    eventKeyAction: -1;
+    tag: number;
+  }
+
+  export type HWKeyEvent =
+    | HWFocusEvent
+    | {
+        eventType: 'up' | 'down' | 'right' | 'left' | string;
+        eventKeyAction: -1 | 1 | 0 | number;
+        tag?: number;
+      };
+
+  export class TVEventHandler {
+    enable<T extends React.Component<unknown>>(
+      component?: T,
+      callback?: (component: T, data: HWKeyEvent) => void
+    ): void;
+    disable(): void;
+  }
+
+  export interface FocusGuideProps extends ViewProps {
+    /**
+     * Array of `Component`s to register as destinations with `UIFocusGuide`
+     */
+    destinations?: (null | number | React.Component<any, any> | React.ComponentClass<any>)[];
+  }
+
+  /**
+   * This component provides support for Apple's `UIFocusGuide` API,
+   * to help ensure that focusable controls can be navigated to,
+   * even if they are not directly in line with other controls.
+   * An example is provided in `RNTester` that shows two different ways of using this component.
+   * https://github.com/react-native-community/react-native-tvos/blob/master/RNTester/js/TVFocusGuideExample.js
+   */
+  export class TVFocusGuideView extends React.Component<FocusGuideProps> {}
+
+  export interface TVTextScrollViewProps extends ScrollViewProps {
+    /**
+     * The duration of the scroll animation when a swipe is detected.
+     * Default value is 0.3 s
+     */
+    scrollDuration?: number;
+    /**
+     * Scrolling distance when a swipe is detected
+     * Default value is half the visible height (vertical scroller)
+     * or width (horizontal scroller)
+     */
+    pageSize?: number;
+    /**
+     * If true, will scroll to start when focus moves out past the beginning
+     * of the scroller
+     * Defaults to true
+     */
+    snapToStart?: boolean;
+    /**
+     * If true, will scroll to end when focus moves out past the end of the
+     * scroller
+     * Defaults to true
+     */
+    snapToEnd?: boolean;
+    /**
+     * Called when the scroller comes into focus (e.g. for highlighting)
+     */
+    onFocus?(evt: HWKeyEvent): void;
+    /**
+     * Called when the scroller goes out of focus
+     */
+    onBlur?(evt: HWKeyEvent): void;
+  }
+
+  export class TVTextScrollView extends React.Component<TVTextScrollViewProps> {}
+}
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "outputDirectory": "reports/junit",
     "outputName": "js-test-results.xml"
   },
+  "types": "index.d.ts",
   "files": [
     "!template/node_modules",
     "!template/package-lock.json",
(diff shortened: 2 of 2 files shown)


dlowder-salesforce (approved):

olafurnsig:
I'm a bit confused here, won't @types/react-native always be overruled by the npm package?

Or would we be importing from 'react-native-tvos' instead?

Krisztiaan:
`@types/react-native` is the original react-native typings, and that is extended by this `.d.ts`. On the top, we're importing it from the `@types` package, so we're not "re-declaring" the whole package, but extending it. You can try it by copying it in your own project with `react-native-tvos` and observing that it "just works"

dlowder-salesforce:
I made a quick test project using https://github.com/react-native-community/react-native-template-typescript ... switched it to use this branch, and it seems to "just work".  (Although the template doesn't seem to come with very strict typechecking, but I'm not really an expert in this area :) )  So I'm cool to merge this, and hopefully the community will let me and you know if there are any problems.

adbl:
@Krisztiaan hmm I'm not getting this to work, it seems I do get the additional types correctly, but everything standard from react-native is not. e.g.:

```
$ yarn run tsc

src/modules/Empty.tsx:2:9 - error TS2305: Module '"react-native"' has no exported member 'View'.

2 import {View} from 'react-native';
```

for `Empty.tsx`
```
import React from 'react';
import {View} from 'react-native';

const Empty: React.FC = () => <View />;
export default Empty;
```

Running typescript 4.0.2. It's possible I have something in `tsconfig.json` causing trouble... any ideas?

adbl:
Oh, when running with `skipLibCheck: false` I'm seeing some things which seem related:

```

node_modules/react-native/index.d.ts:4:3 - error TS2667: Imports are not permitted in module augmentations. Consider moving them to the enclosing external module.

4   import React from 'react';
    ~~~~~~

node_modules/react-native/index.d.ts:4:21 - error TS2307: Cannot find module 'react' or its corresponding type declarations.

4   import React from 'react';
                      ~~~~~~~

node_modules/react-native/index.d.ts:6:3 - error TS2667: Imports are not permitted in module augmentations. Consider moving them to the enclosing external module.

6   import { ViewProps, ScrollViewProps } from '@types/react-native';
    ~~~~~~

node_modules/react-native/index.d.ts:6:46 - error TS6137: Cannot import type declaration files. Consider importing 'react-native' instead of '@types/react-native'.

6   import { ViewProps, ScrollViewProps } from '@types/react-native';
                                               ~~~~~~~~~~~~~~~~~~~~~
```

Also tried with typescript 3.8.3 same problem.

adbl:
Here is a related SO question: https://stackoverflow.com/questions/40456231/typescript-module-augmentation-overwrites-the-original-module

Per the answer, the fix is to change `react-native-tvos/index.d.ts` to:
```
import React from 'react';
import {ViewProps, ScrollViewProps} from 'react-native';

declare module 'react-native' {
...
}
```

But it doesn't see the regular `react-native` type definitions, because `import 'react-native'` is imports itself? I think a solution is to add the types to `@types/react-native-tvos` which augments `@types/react-native`, then include in the README here a reference to it for whoever is using typescript. I guess we'll also have to change our code to say `import ... from `react-native-tvos` in all places which uses TV specific features, annoying.

adbl:
Actually, couldn't we just add the types to `@types/react-native`? There are already things like `TVParallaxProperties` in there.

I can make a PR for that unless anyone disagree? @dlowder-salesforce do you have an opinion?

dlowder-salesforce:
@adbl I agree, that would be simplest.  @Krisztiaan would you be ok with that?

adbl:
Hoping you are ok with it @Krisztiaan! Once the PR gets merged, I'll open another one to remove `index.d.ts` from here.

Krisztiaan:
@dlowder-salesforce Sure thing, I've commented on that PR as well. It's also something I wanted to do but didn't bother.

These errors are easily fixable though in this repository as well. I'm personally not a huge fan of the `DefinitelyTyped` mo.

Krisztiaan:
It should be a simple option to extend the `DefinitelyTyped` definition.
Maybe solutions off the top of my head:
- use `/// <reference types="react-native" />`
- replace `@types/react-native` with `react-native` (TSC _should_ be smart about this, see [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation))

I'm currently working on a SwiftUI project, so if you could try some of the alternatives, that'd be really nice.

I personally would prefer if there'd be an `index.d.ts` in every repository where it's possible (this being one of them), instead of the whole `DefinitelyTyped`. Maybe one day even `react-native` will be `TypeScript`..

adbl:
> @adbl It should be a simple option to extend the `DefinitelyTyped` definition.
> Maybe solutions off the top of my head:
> 
> * use `/// <reference types="react-native" />`

I'll check this out and see if it has any effect, because:

> * replace `@types/react-native` with `react-native` (TSC _should_ be smart about this, see [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation))

this isn't working at the moment, it seems `react-native` resolves to itself due to the package name.

> 
> I'm currently working on a SwiftUI project, so if you could try these options, that'd be really nice.
> 
> I personally would prefer if there'd be an `index.d.ts` in every repository where it's possible (this being one of them), instead of the whole `DefinitelyTyped`. Maybe one day even `react-native` will be `TypeScript`..

adbl:
I tried the following with the same result:

```
/// <reference types="react-native" />

import React from 'react';
import { ViewProps, ScrollViewProps } from '@types/react-native';

declare module 'react-native' {
...
```

I will proceed to try and get it in @types/react-native

Krisztiaan:
I've also encountered @adbl please see #119
As per my local testing, it should work. Circumvents the way typescript recognizes typings.

adbl:
Great, it does indeed work!

EyMaddis:
Very late to the party, but in my project, import `ViewProps` directly no longer works now.

I have problems like 
```
 type MyComProps = ViewProps
 function MyCom(props: MyComProps) {
  return <View {...props} />
}

<MyComp style={...} /> ---> `style does not exist in ....`
```

Following the type definition via Vscode will go to the `ReactNativeTVTypes.d.ts` and only properties like nextFocusDown are available now...

Krisztiaan:
@EyMaddis without looking too deep into it, you may find `https://www.npmjs.com/package/deprecated-react-native-prop-types/` to work to restore the types. Otherwise, the now built in types should be used, and this package updated.

EyMaddis:
Thanks, but this seems to be a general problem. The file also breaks prop types for Touchables for me. I worked around it using `patch-package` to the type to TVViewProps in ReactNativeTVTypes.d.ts.
Might be my setup as I am also overwriting types via react-native-web but I do not really see a reason why that should happen. 
Instead of declaration merging, this just fully overrides the type and Vscode. Very odd.

Krisztiaan:
Unfortunately, it's been a few years since I've worked on a `tvos` project, but when I do pick it up again, this'll be something to look at, unless you want to give it a go, figure out what's up (I'm way out of the loop at the moment), and create a PR, I'd be very happy to review.

PR #126, opened 2020-11-03
Pressable support for TV

### LINKED ISSUE #125: Support the new Pressable API for RN 0.63
React Native 0.63 introduced a new component `Pressable` which is intended as a replacement for `Touchable*`:

https://reactnative.dev/blog/2020/07/06/version-0.63#pressable

I'm starting work on this for TV platforms in the [pressable-for-tv](https://github.com/react-native-tvos/react-native-tvos/tree/pressable-for-tv) branch.  Needs the `isTVSelectable` prop, and needs focus, blur, select handlers.  Once these are working, I'll test `PressableExample` and see what else is required.

dlowder-salesforce:
# Summary

Allow `Pressable` to be functional on Apple TV and Android TV.  Fixes #125 .

- add `focusable` and `isTVSelectable` view props
- add `onFocus()` and `onBlur()`
- export `focused` as a component state as well as `pressed`
- fire `onLongPress()` when Apple TV remote select button gets a long press

## Test Plan

Added examples and made changes in `PressableExample` in RNTester.

### DIFF
--- a/Libraries/Components/Pressable/Pressable.js
+++ b/Libraries/Components/Pressable/Pressable.js
@@ -244,13 +244,9 @@ function Pressable(props: Props, forwardedRef): React.Node {
       if (viewRef.current._nativeTag === evt.target) {
         if (evt?.eventType === 'focus') {
           setFocused(true);
-          setPressed(true);
-          onPressIn && onPressIn(evt);
           onFocus && onFocus(evt);
         } else if (evt.eventType === 'blur') {
           onBlur && onBlur(evt);
-          onPressOut && onPressOut(evt);
-          setPressed(false);
           setFocused(false);
         }
       }
--- a/Libraries/Components/Pressable/Pressable.js
+++ b/Libraries/Components/Pressable/Pressable.js
@@ -28,11 +28,24 @@ import {normalizeRect, type RectOrSize} from '../../StyleSheet/Rect';
 import type {ColorValue} from '../../StyleSheet/StyleSheetTypes';
 import type {LayoutEvent, PressEvent} from '../../Types/CoreEventTypes';
 import View from '../View/View';
+import Platform from '../../Utilities/Platform';
+import typeof TVParallaxPropertiesType from '../AppleTV/TVViewPropTypes';
 
 type ViewStyleProp = $ElementType<React.ElementConfig<typeof View>, 'style'>;
 
 export type StateCallbackType = $ReadOnly<{|
   pressed: boolean,
+|}>;
+
+type TVProps = $ReadOnly<{|
+  hasTVPreferredFocus?: ?boolean,
+  isTVSelectable?: ?boolean,
+  tvParallaxProperties?: ?TVParallaxPropertiesType,
+  nextFocusDown?: ?number,
+  nextFocusForward?: ?number,
+  nextFocusLeft?: ?number,
+  nextFocusRight?: ?number,
+  nextFocusUp?: ?number,
 |}>;
 
 type Props = $ReadOnly<{|
@@ -131,6 +144,10 @@ type Props = $ReadOnly<{|
    * Used only for documentation or testing (e.g. snapshot testing).
    */
   testOnly_pressed?: ?boolean,
+  /**
+   * Props needed for Apple TV and Android TV
+   */
+  ...TVProps,
 |}>;
 
 /**
@@ -146,6 +163,7 @@ function Pressable(props: Props, forwardedRef): React.Node {
     delayLongPress,
     disabled,
     focusable,
+    isTVSelectable,
     onLongPress,
     onPress,
     onPressIn,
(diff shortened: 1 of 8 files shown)


TheSavior:
This looks reasonable although @rectified95 might be a better reviewer having implemented this for windows.

As a meta comment, I’m surprised by the usage of findNodeHandle. Does that end up getting passed through to a native module or something? We are trying to get rid of findNodeHandle so it would be good to know if your use case has a different solution

dlowder-salesforce:
The focus/blur events have the React tag has one of their properties.  FindNodeHandle is used to make sure the event's tag matches the React tag for the Pressable.

I'll do some testing to make sure we really need that check.

Libraries/Components/Pressable/Pressable.js
```diff
@@ -210,17 +240,44 @@ function Pressable(props: Props, forwardedRef): React.Node {
   );
   const eventHandlers = usePressability(config);
 
+  const pressableTVEventHandler = (evt: Event) => {
+    if (props.isTVSelectable !== false || props.focusable !== false) {
+      const tag = findNodeHandle(viewRef.current);
+      if (tag === evt.tag) {
```
rectified95, line 246: Disclaimer: I don't know anything about the TV SDK.

If a given Pressable has received an event, why do you need to check if the tags match? Is that because you're using the useTVEventHandler hook? 
Afaik, the way event dispatching works is that a view manager includes the tag in the event payload, and RN funnels it to  the correct node, so eg. if you press a Pressable, you don't need to check the tag in the incoming event, but rather the onPress callback gets invoked accordingly. 
Which ViewManager does the focus/blur event come from here? On Windows it's the ViewViewManager, so we just hook up the onFocus stuff to the View that Pressable wraps (right below this function, in render()).

RNTester/js/examples/Pressable/PressableExample.js
```diff
@@ -54,6 +54,34 @@ function ContentPress() {
...
+  if (timesPressed > 1) {
+    textLog = timesPressed + 'x onPress';
+  } else if (timesPressed > 0) {
+    textLog = 'onPress';
+  }
+
+  return (
+    <>
+      <View style={styles.row}>
+        <Pressable
```
rectified95, line 70: might be useful to actually show the usage of onFocus/onBlur in the example here - eg.:
onBlur={() => console.log('Pressable onBlur')} or sth making  a change to the UI.

Libraries/Components/Pressable/Pressable.js
TheSavior, line 246: Ah, knowing that you are just checking for whether it matches, if you find you need this, we added `event.target` that is a ref to the HostComponent.

You should be able to do: `if (viewRef.current === evt.target)`

Check out ScrollResponder for an example: https://github.com/facebook/react-native/blob/master/Libraries/Components/ScrollResponder.js#L190

TheSavior: This change went out as part of 0.63: https://github.com/facebook/react-native/commit/6286270e4cb10b40cfd7c8193e31d965f6815150

rectified95 (commented):
Overall LGTM, uses the same mechanism of exposing onFocus from Pressability as Windows.
Would add more examples of the new functionality, though.

Libraries/Components/Pressable/Pressable.js
dlowder-salesforce: That's fantastic.... I'll do an audit of some other components too to see where we can convert to using evt.target

dlowder-salesforce: @rectified95 your intuition is correct... the useTVEventHandler hook doesn't keep info on the component that fired the event.  This doesn't matter for remote control events (up/down/left/right/select) but it does matter for focus/blur.  Now that @TheSavior has provided me the correct APIs for 0.63, I can go and fix useTVEventHandler and make Pressable work correctly.

rectified95 (approved):

dlowder-salesforce:
OK things are looking good now.  I'm still not sure if it makes sense to have both `pressed` and `focused` since they are really doing the same thing at the moment.  It might make more sense to call `onPressIn`, `onPress`, and `onPressOut` successively when the TV remote is clicked, and use only the focus props when handling focus/blur.

TheSavior:
I would expect the dpad on the remote to change focus (focused) of different elements, and then to render with pressed when the select button is pressed. That would be necessary for rendering a different UI on selection vs a scale animation on press.

dlowder-salesforce:
@TheSavior ok I've made the change to separate out the behavior of `pressed` and `focused`, and updated the README.  Things look good on both Android TV and Apple TV simulators.  TV parallax animations work on Apple TV, and the custom ripples work on Android TV.

TheSavior:
Looks good to me. I'm not sure if there is a distinction between hovered and focused on TV systems, but we expect Pressable to eventually have `{pressed, focused, hovered}`. As you know TV best you should decide how you want this platform to respond with either focused or hovered.

dlowder-salesforce:
Thanks @TheSavior .... I don't think `hovered` has any meaning on a TV platform, but if it is eventually there in RN core `Pressable` then we'll figure out how to handle it.

Issue #128, opened 2020-11-12
[androidTV] Scrollview still scrollable with dpad when scrollEnabled={false} #29774


prakash-altair:
Description
A scrollview with the scrollEnabled prop set to false still scrolls on tv remote dpad input on androidTV.
Also onScroll events are fired when the list is scrolled this way.

React Native version:
  ##0.63.1-2

Expected Results
The scrollview should not scroll when scrollEnabled is set to false, regardless of touch, mouse or dpad input. 
facebook/react-native#29774

prakash-altair:
[https://github.com/facebook/react-native/issues/29774](url)

prakash-altair:
`  @ReactProp(name = "scrollEnabled", defaultBoolean = true)
  public void setScrollEnabled(ReactHorizontalScrollView view, boolean value) {
    view.setScrollEnabled(value);
    
    // Set focusable to match whether scroll is enabled. This improves keyboarding
    // experience by not making scrollview to scroll when scroll enabled is to false.
    view.setFocusable(value);
  }`

Adding `view.setFocusable(value);` to `ReactHorizontalScrollViewManager.java` will solve this issue.  Should i rise a PR for this to be added in the next release ?

dlowder-salesforce:
Yes please send a PR for this!!! 😊

prakashjais99:
#130

prakashjais99:
@dlowder-salesforce 
Thanks for merging PR

dlowder-salesforce:
@prakashjais99 no problem, thanks for the PR!  Once #129 is resolved, I'll do a new release.

dlowder-salesforce:
Fixed in 0.63.1-4 release

PR #130, opened 2020-11-16
Fix for scrolling issue #128

prakashjais99:
Fix horizontal scroll view scrolling using DPad even when scrollEnabled is set to false

## Summary

A scrollview with the scrollEnabled prop set to false still scrolls on tv remote dpad input on androidTV.
This effect can be seen when a Vertical ScrollView with scrollEnabled={false} have multiple HorizontalScrollView with scrollEnabled={false}. When Dpad up down is pressed HorizontalScrollview which is present inside Vertical ScrillView moves like listview.

React Native version:
##0.63.1-2

Expected Results
The scrollview should not scroll when scrollEnabled is set to false, regardless of touch, mouse or DPad input.

## Changelog

This change is continuation of PR  [#25105](https://github.com/facebook/react-native/pull/25105) which had handled scroll events for vertical scroll view but missed to had changes in `ReactHorizontalScrollViewManager.java`

[Android] [TV] - HorizontalScrollView

## Test Plan

Horizontal scroll view should scroll when scrollEnabled ={true} and should not scroll when set to false  regardless of touch, mouse or dpad input.


prakashjais99:
@dlowder-salesforce can we have this merged in the next release as we are blocked on this and using fork of repo now

PR #214, opened 2021-08-24
Focus guide exception fix

AnkitRathi7:
## Summary

TVFocusGuideView's reactSuperView is nil

## Changelog

I observed that in some cases, in first render, `reactSuperview` of `TVFocusGuideView` is nil which causes and exception with constraint. adding check for nil superview fixes the issue.
FocusGuide-constraint-issue

[CATEGORY] [TYPE] - Message

### DIFF
--- a/React/Views/RCTTVFocusGuideView.m
+++ b/React/Views/RCTTVFocusGuideView.m
@@ -44,7 +44,7 @@
 - (void)addFocusGuide:(NSArray*)destinations {
   
   UIView *origin = [self reactSuperview];
-  if (self.focusGuide == nil) {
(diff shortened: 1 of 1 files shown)


douglowder:
@AnkitRathi7 thanks very much for this contribution! Is there any pattern to the cases where this happens? Can you reliably reproduce the issue on an example project?

douglowder (approved):

AnkitRathi7:
@douglowder Thanks for merging. While I am not able clearly identify why reactSuperview is nil, I see this issue mostly when  there is an initial value of `destinations` instead of empty array.

Issue #216, opened 2021-08-30
UI Layout is incorrect whenever we use TVFocusGuideView in tvOS projects.


Ganesh-Kumar-Reddy-M:
## Description
The **TVFocusGuideView** uses two views internally to properly set up the **FocusGuide** and it accepts the **'style'** prop along with other view props. But the problem is that, the **TVFocusGuideView** applies the same 'style' prop to both same props to both parent and child views, which is causing the UI issue.
For example, if the client specifies **'marginLeft'** property in the **style** prop, then the **TVFocusGuideView** will apply the **'marginLeft'** for both parent and child, and so, left margin is getting added twice and UI becomes incorrect. The same is applicable for other layout properties.

## React Native version:
react-native-tvos@0.64.2-2

## Steps To Reproduce
1. Use **TVFocusGuideView** in ReactNative tvOS project.
2. Specify the layout related properties like **'marginLeft'**, **'marginTop'** in the **'style'** prop and pass it to the **TVFocusGuideView**.
3. Now observe layout properties will be applied twice and UI looks incorrect.

## Expected Results
UI should be should be laid out as specified in the 'style' prop.

## Code example:
<TVFocusGuideView
          destinations={destinations}
          style={{marginLeft: 20, marginTop: 20, width: 600, height: 200}}>
         .....
 </TVFocusGuideView>

douglowder:
I have one other issue I'm working on at the moment -- once I've merged the fix for that as well, I will create a release and close this.

douglowder:
Fixed in 0.64.2-3.

PR #217, opened 2021-08-30
Fix the issue #216: tvOS - UI Layout is incorrect for TVFocusGuideView.

Ganesh-Kumar-Reddy-M:
UI Layout is incorrect whenever we use TVFocusGuideView in tvOS projects. For details, see #216 

## Summary
The TVFocusGuideView is using two views internally to properly set up the FocusGuide. But the problem is that, if the client specifies any layout related properties in the 'style' prop, TVFocusGuideView will apply the same props to both parent and child views, which is causing the UI issue.
For example, if the client specifies 'marginLeft' property in the style prop, then the TVFocusGuideView will apply the 'marginLeft' for both parent and child, and so, left margin is getting added twice and UI becomes incorrect. The same is applicable for other layout properties.

**React Native version:**
react-native-tvos@0.64.2-2

## Changelog

[tvOS] [Fixed] - Fix UI issue in **TVFocusGuideView**

## Test Plan

UI Layout should be correct for FocusGuide even when client passed layout related properties like **'marginLeft'**, **'marginRight'**, in the **'style'** prop to the **TVFocusGuideView**

### DIFF
--- a/Libraries/Components/AppleTV/TVFocusGuideView.js
+++ b/Libraries/Components/AppleTV/TVFocusGuideView.js
@@ -42,9 +42,19 @@ class TVFocusGuideView extends React.Component<FocusGuideProps> {
     return (
       // Container view must have nonzero size
       <ReactNative.View style={[{minHeight: 1, minWidth: 1}, this.props.style]}>
+        {
+          /**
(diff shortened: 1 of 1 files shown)


douglowder:
Thanks for this fix! I'll test it out locally and maybe add your repro for this in `packages/rn-tester` as well.

Issue #233, opened 2021-10-29
TVEventHandler and useTVEventHandler captured event repetition


ali-sao:
## Description
Hello and thank you for the support; 
I am building a sliding component using TVEventHandler on **Android TV** when I noticed that a single event is logged twice, I used the useTVEventHandler hook for the same purpose, same side effect. 

<img width="966" alt="Screen Shot 2021-10-29 at 8 57 11 PM" src="https://user-images.githubusercontent.com/6895166/139489359-35ae23dd-33ab-41cf-b2f8-7c7d8ec914d5.png">

## React + Native /Deps info:
`System:
    OS: macOS 11.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 299.29 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.12.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 8.1.0 - /usr/local/bin/npm
    Watchman: 2021.10.18.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.2 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 15.0, DriverKit 20.4, macOS 11.3, tvOS 15.0, watchOS 8.0
    Android SDK:
      API Levels: 28, 29, 30
      Build Tools: 28.0.3, 29.0.2, 29.0.3, 30.0.2, 30.0.3
      System Images: android-Q | Android TV Intel x86 Atom
      Android NDK: 22.1.7171670
  IDEs:
    Android Studio: 2020.3 AI-203.7717.56.2031.7621141
    Xcode: 13.0/13A233 - /usr/bin/xcodebuild
  Languages:
    Java: 14.0.1 - /Library/Java/JavaVirtualMachines/jdk-14.0.1.jdk/Contents/Home/bin/javac
  npmPackages:
    react: 17.0.1 => 17.0.1
    react-native-tvos:  0.64.2-4`

## Steps To Reproduce
use the example mentioned in this repo readme on a boilerplate app. [https://github.com/react-native-tvos/react-native-tvos#code-changes](url)

## Expected Results
To log the event once. 

## Code example
`import React from "react";
  import {View,Text,useTVEventHandler,TouchableOpacity} from "react-native";

const App =() => {
[...]

PR #258, opened 2022-01-23
Android TV: feature flag for key up events (#233) (#251)

### LINKED ISSUE #251: Android TV: create a prop that controls firing of keyDown events
keyDown events were removed to fix the issue where each key press resulted in two events surfaced in TVEventHandler on Android.

However, this breaks certain use cases for some TV developers.  See discussion in #233 .

We need a prop for Android that controls whether keyDown events are fired or not.

douglowder:
Adding `enableKeyDownEvents` in `ReactFeatureFlags` class.  If this is set in your application startup code, both key down and key up events will be sent to TVEventHandler.  See discussion in related issue #233 . Fixes #251.

Example from template application:

```java
  @Override
  public void onCreate() {
    super.onCreate();
    SoLoader.init(this, /* native exopackage */ false);
    // Normally we only send key up events in ReactAndroidHWInputDeviceHelper
    // Change enableKeyDownEvents to true to send both key down and key up events
    ReactFeatureFlags.enableKeyDownEvents = false;
```

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/config/ReactFeatureFlags.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/config/ReactFeatureFlags.java
@@ -89,4 +89,9 @@ public class ReactFeatureFlags {
   public static boolean enableLockFreeEventDispatcher = false;
 
   public static boolean enableAggressiveEventEmitterCleanup = false;
+
+  /**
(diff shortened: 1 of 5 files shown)


etiennelab:
@douglowder it seems OK to me. Thank you !

PR #259, opened 2022-01-24
Implement single event listener for Touchable components to boost performance

Zahoq:
## Summary

Each Touchable component needlessly creates own event listener, which, after focus transition, results in multiple handlers called just to check if event should be even handled by particular component. Performance of focus transition can be improved.

## Changelog

[General] [Changed] - Improved focus transition performance of Touchable components by introducing single Event Listener for all Touchables, instead of one Event Listener per Touchable.

## Test Plan

I've prepared simple TV app with improvement implemented as a patch to check and measure performance improvement. Rough measurement of focus transition between tiles on AndroidTV emulator indicates that time between onBlur -> onFocus pair reduced from ~11 ms to ~7 ms on average, and peak value form ~48 ms to ~14 ms.
https://github.com/Zahoq/TouchableFocusImprovementTestApp

### DIFF
--- a/Libraries/Components/AppleTV/TVFocusEventHandler.js
+++ b/Libraries/Components/AppleTV/TVFocusEventHandler.js
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+import NativeEventEmitter from '../../EventEmitter/NativeEventEmitter';
+import Platform from '../../Utilities/Platform';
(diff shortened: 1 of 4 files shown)


douglowder:
This looks great. Could you also modify `Libraries/Components/Pressable/Pressable.js` to use the new focus handler?

douglowder (changes requested):
This looks great. Could you also modify `Libraries/Components/Pressable/Pressable.js` to use the new focus handler?

Zahoq:
> This looks great. Could you also modify `Libraries/Components/Pressable/Pressable.js` to use the new focus handler?
Sure. I'll move initialization of `TVFocusEventHandler` instance to `TVFocusEventHandler.js` so both Touchables and Pressables use same instance of that class.

douglowder (approved):
LGTM, tested fine in simulator

PR #266, opened 2022-03-01
Add support for Page Up / Page down (channel buttons) on iOS remote control

BryanRolfe:
## Summary

This adds support for the Channel Up/Down buttons on the iOS control center remote control.

## Changelog

[TVOS] [Added] - Add support for Channel Up/Down buttons on iOS control center remote control.

## Test Plan

Add a TV event handler that logs eventType, open the Apple TV remote on an iPhone and press channel up/down to see the new pageUp/Down events


BryanRolfe:
Broke my commit and missed a few lines of code

Issue #291, opened 2022-04-06
Last focused view's reference is held even after it's detached from window


DrRefactor:
### Description

RCTRootViewInternal.h helds strong reference to preferred focus view:

```
@property (nonatomic, strong) UIView *reactPreferredFocusedView;
```

which is saved each time hasTVPreferredFocus is used to focus element (RCTTVView.m):

```
- (void)setHasTVPreferredFocus:(BOOL)hasTVPreferredFocus
{
  _hasTVPreferredFocus = hasTVPreferredFocus;
  if (hasTVPreferredFocus) {
    dispatch_async(dispatch_get_main_queue(), ^{
      UIView *rootview = self;
      while (![rootview isReactRootView] && rootview != nil) {
        rootview = [rootview superview];
      }
      if (rootview == nil)
        return;
      
      rootview = [rootview superview];
      
      [(RCTRootView *)rootview setReactPreferredFocusedView:self]; /// <------ THIS LINE
      [rootview setNeedsFocusUpdate];
      [rootview updateFocusIfNeeded];
    });
  }
}
```

Due to reactPreferredFocusedView being strong reference, it's not released even if element unmounts.
This leads to bugs in my code, as I base some of native focus management on native code, where I store last focused element as a weak reference to UIView, to be able to restore remembered focus in given views.

I obviously can verify, whether remembered UIView is attached to window (view.window == nil), and ignore it if not. However, holding strong reference to detached UIView still seems like a bug and a minor memory leak.

### Version

0.66.3-1

### Output of `npx react-native info`

System:
    OS: macOS 12.0.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 33.57 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.2 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 8.1.2 - /usr/local/bin/npm
    Watchman: 2022.02.14.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /Users/CENSORED/.rbenv/shims/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.2, iOS 15.2, macOS 12.1, tvOS 15.2, watchOS 8.3
    Android SDK:
      API Levels: 28, 29, 30, 31
[...]

PR #292, opened 2022-04-06
Drrefactor/root focused view leak

DrRefactor:
## Summary
Please see issue with detailed description:
https://github.com/react-native-tvos/react-native-tvos/issues/291

## Changelog
Fix potential focus race related to detached focused view's reference being held in RCTRootView.

[Internal] [Fixed] - Prevent using detached RCTView as preferredFocusEnvironment

### DIFF
--- a/React/Base/RCTRootView.m
+++ b/React/Base/RCTRootView.m
@@ -168,7 +168,7 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder : (NSCoder *)aDecoder)
 #if TARGET_OS_TV
 - (NSArray<id<UIFocusEnvironment>> *)preferredFocusEnvironments
 {
-  if (self.reactPreferredFocusedView) {
+  if (self.reactPreferredFocusedView && self.reactPreferredFocusedView.window != nil) {
     return @[self.reactPreferredFocusedView];
   }
   return [super preferredFocusEnvironments];
--- a/React/Base/RCTRootViewInternal.h
+++ b/React/Base/RCTRootViewInternal.h
@@ -26,7 +26,7 @@
  */
 #if TARGET_OS_TV
 @property (nonatomic, strong) RCTTVRemoteHandler *tvRemoteHandler;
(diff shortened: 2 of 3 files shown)


React/Base/RCTRootView.m
```diff
@@ -168,7 +168,7 @@ - (RCTModuleRegistry *)moduleRegistry
 #if TARGET_OS_TV
 - (NSArray<id<UIFocusEnvironment>> *)preferredFocusEnvironments
 {
-  if (self.reactPreferredFocusedView) {
+  if (self.reactPreferredFocusedView && self.reactPreferredFocusedView.window != nil) {
```
DrRefactor, line 171: Unfortunately, for short moment after view unmount, this reference is still not released, therefore additional check is needed.

DrRefactor: This need to be verified in case of some custom views, like modals, maybe it makes sense for unattached view, to be preferredFocusEnvironment in some cases?

React/Base/RCTRootViewInternal.h
```diff
@@ -26,7 +26,7 @@
  */
 #if TARGET_OS_TV
 @property (nonatomic, strong) RCTTVRemoteHandler *tvRemoteHandler;
-@property (nonatomic, strong) UIView *reactPreferredFocusedView;
+@property (nonatomic, weak) UIView *reactPreferredFocusedView;
```
DrRefactor, line 29: before this fix, UIView's reference was held even 30 seconds after unmount. I haven't tested anything above, assuming implication: 30s => will be there forever

douglowder:
Thanks for the PR! I'll test this out. If we are going to do this, we would need a similar change in the view class for the new Fabric architecture.

DrRefactor:
I applied second commit, regarding detached views, to Fabric RCTRootComponentView.

I see that you already use weak reference in RCTRootComponentView.h -> reactPreferredFocusedView, so this one is already done.

douglowder (approved):

PR #300, opened 2022-04-16
Enhancement to the nextFocus APIs for TV

douglowder:
At the moment, the nextFocusLeft/Right/Up/Down APIs need React view tags to be passed in, so the app has to use findNodeHandle() to get the tags. This change allows component refs to be passed in directly, and the tags are calculated internally before passing to native layer.  Passing in tags directly will still work, so this is not a breaking change.

This PR also includes:

- Modified `DirectionalNextFocusExample` to demonstrate passing in component refs to nextFocus APIs
- Refactor `DirectionalNextFocusExample` and `TVFocusGuideViewExample` to use modern functional components

### DIFF
--- a/Libraries/Components/AppleTV/tagForComponentOrHandle.js
+++ b/Libraries/Components/AppleTV/tagForComponentOrHandle.js
@@ -1,0 +1,34 @@
+// Replacement for findNodeHandle, since we need this for converting
+// components to tags when rendering nextFocus props
+// @flow
+
+// TODO: make this work for Fabric
+
+// import {findNodeHandle} from '../../Renderer/shims/ReactNative';
+
+type TagForComponentOrHandleType = (
+    component: ?(
+      | {_nativeTag: ?number, canonical: ?{_nativeTag: ?number}}
+      | number
+    ),
+  ) => ?number;
+
+const tagForComponentOrHandle: TagForComponentOrHandleType = (
+  component: ?(
+    | {_nativeTag: ?number, canonical: ?{_nativeTag: ?number}}
+    | number
+  ),
+): ?number => {
+  if (typeof component === 'number') {
+    return component;
+  }
+  if (typeof component?._nativeTag === 'number') {
+    return component?._nativeTag;
+  }
+  if (typeof component?.canonical?._nativeTag === 'number') {
+    return component?.canonical?._nativeTag;
+  }
+  return undefined;
+};
+
+export default tagForComponentOrHandle;
--- a/Libraries/Components/Touchable/TouchableOpacity.js
+++ b/Libraries/Components/Touchable/TouchableOpacity.js
@@ -20,6 +20,7 @@ import type {ViewStyleProp} from '../../StyleSheet/StyleSheet';
 import flattenStyle from '../../StyleSheet/flattenStyle';
 import Platform from '../../Utilities/Platform';
 import typeof TVParallaxPropertiesType from '../AppleTV/TVViewPropTypes';
+import tagForComponentOrHandle from '../AppleTV/tagForComponentOrHandle';
 
 import * as React from 'react';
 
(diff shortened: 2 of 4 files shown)


freeboub:
Just have a quick look into this PR, this file is located in AppleTV folder but is generic to all devices...
[tagForComponentOrHandle.js](https://github.com/react-native-tvos/react-native-tvos/pull/300/files#diff-fb755fcbe5c491f3dc716793dd822d6c38c8a8345402fcc1779ccedb371f84c7)

I think this is a good feature, but shouldn't it be pushed in the main react-native repository ?
Thank you

douglowder:
@freeboub agreed, I will create a separate issue to track that work after this is merged.

Libraries/Components/Touchable/TouchableOpacity.js
```diff
@@ -245,13 +246,15 @@ class TouchableOpacity extends React.Component<Props, State> {
...
-        nextFocusUp={this.props.nextFocusUp}
+        nextFocusDown={tagForComponentOrHandle(this.props.nextFocusDown)}
+        nextFocusForward={tagForComponentOrHandle(this.props.nextFocusForward)}
+        nextFocusLeft={tagForComponentOrHandle(this.props.nextFocusLeft)}
+        nextFocusRight={tagForComponentOrHandle(this.props.nextFocusRight)}
+        nextFocusUp={tagForComponentOrHandle(this.props.nextFocusUp)}
         hasTVPreferredFocus={this.props.hasTVPreferredFocus === true}
-        isTVSelectable={this.props.isTVSelectable !== false && this.props.accessible !== false}
+        isTVSelectable={
+          this.props.isTVSelectable !== false && this.props.accessible !== false
```
Krisztiaan, line 256: Unrelated, but this may be unexpected check, when 0 or "" is passed to these `boolean`s, maybe remove the strict equation check?

Libraries/Components/AppleTV/tagForComponentOrHandle.js
```diff
@@ -0,0 +1,34 @@
+// Replacement for findNodeHandle, since we need this for converting
+// components to tags when rendering nextFocus props
+// @flow
+
+// TODO: make this work for Fabric
+
+// import {findNodeHandle} from '../../Renderer/shims/ReactNative';
```
Krisztiaan, line 7: What's the reason for not use the original? Seems to do the same thing and some more.
https://github.com/facebook/react-native/blob/41cbccd98dd6c98d1f662674164cf455105a1359/Libraries/Renderer/implementations/ReactNativeRenderer-prod.js#L8067

Libraries/Components/Touchable/TouchableOpacity.js
```diff
@@ -245,13 +246,15 @@ class TouchableOpacity extends React.Component<Props, State> {
         nativeID={this.props.nativeID}
         testID={this.props.testID}
         onLayout={this.props.onLayout}
-        nextFocusDown={this.props.nextFocusDown}
-        nextFocusForward={this.props.nextFocusForward}
-        nextFocusLeft={this.props.nextFocusLeft}
-        nextFocusRight={this.props.nextFocusRight}
-        nextFocusUp={this.props.nextFocusUp}
+        nextFocusDown={tagForComponentOrHandle(this.props.nextFocusDown)}
```
Krisztiaan, line 249: Could be nice, at the time I was also thinking this could be done via a11y labels / ids, as is recommended for when one wants to look up components for tests. That'd also promote better a11y for lib users.

Krisztiaan (commented):
Not entirely sure the API was great for starters. Looks good anyways.

Wondering about `tagForComponentOrHandler` too, as it seems like RN already has that.

Libraries/Components/Touchable/TouchableOpacity.js
douglowder: @Krisztiaan thanks for the review!

- RN does already have `findNodeHandle`, however it is not allowed to be used in this context and will throw an error. I created my own method here and extracted it to its own file, and it will continue to change as the new architecture is rolled out.
- Those booleans have default values and should not be set to anything other than true/false, so I'd like to leave the strict equation checks :)

Krisztiaan (approved):
Looks good to me!

PR #301, opened 2022-04-19
update tvos-types to include pan events

bwallberg:
## Summary

The TVMenuControl was renamed to TVEventControl & 'pan' events were added

## Changelog

- Remove un-used `HWFocusEvent` since in practise the type-system could only ever reach the "common" event
- Renamed TVMenuContro to TVEventControl
- Added the missing fields for the 'pan' events
- Renamed HWKeyEvent to HWEvent since it contextuelly made more sense based on recent changes

### DIFF
--- a/tvos-types.d.ts
+++ b/tvos-types.d.ts
@@ -31,31 +31,32 @@ declare module 'react-native' {
   nextFocusUp?: number,
   }
 
-  export const useTVEventHandler: (handleEvent: (event: HWKeyEvent) => void) => void;
+  export const useTVEventHandler: (handleEvent: (event: HWEvent) => void) => void;
 
-  export const TVMenuControl: {
+  export const TVEventControl: {
     enableTVMenuKey(): void;
     disableTVMenuKey(): void;
+    enableTVPanGesture(): void;
+    disableTVPanGesture(): void;
   };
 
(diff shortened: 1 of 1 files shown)


tvos-types.d.ts
```diff
@@ -31,31 +31,32 @@ declare module 'react-native' {
...
-  export type HWKeyEvent =
-    | HWFocusEvent
-    | {
-        eventType: 'up' | 'down' | 'right' | 'left' | string;
-        eventKeyAction: -1 | 1 | 0 | number;
-        tag?: number;
-      };
+  export type HWEvent = {
+    eventType: 'up' | 'down' | 'right' | 'left' | 'blur' | 'focus' | 'pan' | string;
+    eventKeyAction: -1 | 1 | 0 | number;
```
douglowder, line 45: eventKeyAction should be optional (Apple TV doesn't send this)

douglowder (changes requested):
Looks good! Needs one small change

tvos-types.d.ts
bwallberg: Fixed! 

I also noticed that I had the wrong casing on the body state

douglowder (approved):

PR #302, opened 2022-04-19
add focused prop to Pressable API's PressableStateCallbackType

joeflateau:
## Summary

react-native-tvos adds a "focused" prop to Pressable's callback props (style and children). This PR adds that prop to the type definitions so it can be used in typescript code. 

## Changelog

[General] [Added] - Pressable's focused callback prop types

## Test Plan

```tsx
          <View>
              <Pressable>
                {({focused}) => (
                  <Text
                    style={tw`${focused ? 'text-red-500 ' : 'text-blue-500 '}`}>
                    Thing 1
                  </Text>
                )}
              </Pressable>
              <Pressable>
                {({focused}) => (
                  <Text
                    style={tw`${focused ? 'text-red-500 ' : 'text-blue-500 '}`}>
                    Thing 2
                  </Text>
                )}
              </Pressable>
              <Pressable>
                {({focused}) => (
                  <Text
                    style={tw`${focused ? 'text-red-500 ' : 'text-blue-500 '}`}>
                    Thing 3
                  </Text>
                )}
              </Pressable>
            </View>
```

### DIFF
--- a/tvos-types.d.ts
+++ b/tvos-types.d.ts
@@ -111,4 +111,8 @@ declare module 'react-native' {
   }
 
   export class TVTextScrollView extends React.Component<TVTextScrollViewProps> {}
+  
+  export interface PressableStateCallbackType {
+    readonly focused: boolean;
+  }
 }


tvos-types.d.ts
```diff
@@ -111,4 +111,8 @@ declare module 'react-native' {
   }
 
   export class TVTextScrollView extends React.Component<TVTextScrollViewProps> {}
+  
+  export interface PressableStateCallbackType {
+    readonly focused: boolean;
```
douglowder, line 116: Does this overwrite the Typescript type for the `pressed` state, which presumably already exists in React Native core types?

douglowder (changes requested):
One question inline

tvos-types.d.ts
joeflateau: nope! they are merged, not overwritten: https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces the result is that PressableStateCallbackType has both focused and pressed properties.

douglowder (approved):

Issue #206, opened 2021-08-06
Proposal : TVFocusGuideView for Android


prakashjais99:
Currently, it's very difficult to direct Focus between Viewgroups of Touchables 

**ForExample** 
1) In the below image it's very difficult to direct focus from any items in cluster 2 to the previously focused item in cluster 1 to achieve this currently we have to set `nextFocusUp` props for all the children in cluster 2 to cluster 1 last focused item id.

<img width="960" alt="tv-nav-focusable-1" src="https://user-images.githubusercontent.com/5467335/128498686-fc097094-352e-4938-baa6-011d0a199dbd.png">

2) In the below example from all the horizontal list items we should open a previously selected menu, which is the second item. For this, we have to set all the first items of lists `nextFocusLeft` to the 2nd menu item id 

<img width="247" alt="Screenshot 2021-08-06 at 4 19 12 PM" src="https://user-images.githubusercontent.com/5467335/128499553-6e998fe4-10fd-4956-b869-80067cc9e807.png">

Also, one more problem is we have to pass node id between components to achieve this via props which make components reusability an issue

**Solution**
If we have WrapperView like TVFocusFocusGuide which gets focus first for a cluster and then based on requirement we can pass the focus to any of its children's.

We have  `android:descendantFocusability`  and `onRequestFocusInDescendants` view group props which will let us achieve this behavior. 

We can have a component that wraps touchable and controls the focus setup and movement between view groups.

douglowder:
This is a great idea, I wonder if we can extend `TVFocusGuideView` to have this behavior on Android.....

prakashjais99:
We can do like below
1) Set the FocusGuideView to get the focus in the child View prior to it
```
setDescendantFocusability(ViewGroup.FOCUS_BEFORE_DESCENDANTS);
this.setFocusable(true);
```
2) Record the focused view in the requestChildFocus method
```
@Override
public void requestChildFocus(View child, View focused) {
       super.requestChildFocus(child, focused);
    if (null != child) {
        mLastFocusViewID = child.getId()
    }
}
```
3) When FocusGuideView is about to get focus, actively set the last focused View to get focus
```
@Override
public boolean requestFocus(int direction, Rect previouslyFocusedRect) {
      View lastFocusedView = getLayoutManager().findViewById(mLastFocusViewID);
      if(lastFocusedView != null) {
             lastFocusedView.requestFocus();
             return false;
     }
}
```

[...]

PR #313, opened 2022-06-14
Feature/tvfocusguideview android

SebastienChauvin:
## Summary

Add support for TVFocusGuideView on Android: https://github.com/react-native-tvos/react-native-tvos/issues/206

## Changelog

Android Added - Support for TVFocusGuideView

## Test Plan

Tested with rn-tester / TVFocusGuideView

https://user-images.githubusercontent.com/7773037/173526216-7d1199c7-d532-4182-b551-9bcc23a540f4.mov

### DIFF
--- a/Libraries/Components/AppleTV/TVFocusGuideView.js
+++ b/Libraries/Components/AppleTV/TVFocusGuideView.js
@@ -11,9 +11,10 @@
 const React = require('react');
 const ReactNative = require('react-native');
 import ReactNativeShims from '../../Renderer/shims/ReactNative';
-const Platform = require('../../Utilities/Platform');
 import {Commands} from '../../Components/View/ViewNativeComponent';
 import type {ViewProps} from '../View/ViewPropTypes';
+
+const Platform = require('../../Utilities/Platform');
 
 type FocusGuideProps = $ReadOnly<{
   ...ViewProps,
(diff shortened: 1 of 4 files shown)


Libraries/Components/AppleTV/TVFocusGuideView.js
```diff
@@ -25,42 +26,64 @@ type FocusGuideProps = $ReadOnly<{
 }>;
 
 const TVFocusGuideView = (props: FocusGuideProps) => {
-  const focusGuideRef = React.useRef(null);
+  const tvOSFocusGuideRef = React.useRef(null);
+  const [nativeDestinations, setNativeDestinations] = React.useState([]);
 
   React.useEffect(() => {
-    const nativeDestinations = (props.destinations || [])
```
douglowder, line 31: It looks like your changes are based on an earlier version of this repo. In particular, your changes would revert https://github.com/react-native-tvos/react-native-tvos/commit/6896ad6d08cc2a2bd237c9486e0f5702ad03ffea and break tvOS.

douglowder (changes requested):
See inline comment

Libraries/Components/AppleTV/TVFocusGuideView.js
SebastienChauvin: In my tests, tvOS was still working. I will check again the code changes.

SebastienChauvin: After looking at this again, the behaviour has not changed for tvOS. the state has been an introduced as an intermediary but ultimately Commands.setDestination is called in the same way as before.

douglowder: Have you tested your changes when running in the new Fabric architecture?

SebastienChauvin: Yes  
```
pod install          
Building RNTester with Fabric enabled.
```
--> Same behaviour
And again, the state is only there to share code between Android and tvOS. But the tvOS does the same native command (see the second useEffect)

douglowder: Thank you!!! I do see that things are working fine in local testing, so I'll go ahead and merge this.

douglowder (approved):

PR #314, opened 2022-06-14
fix: add default padding to focus guide view to fix focus order with its children

leonchabbey:
## Summary

We are having issues when a TVFocusGuideView has children which edges are the same are the parent. It usually happens when there are no padding on the parent and no margin on the children.

### **Without padding** --> The TVFocusGuideView is sometimes (most of the time) ignored and the focus goes to the child which is also accessible in the focus direction.
https://user-images.githubusercontent.com/12344434/173606404-1427dc98-453c-401e-81e6-7c8eb222da3f.mov

### **With `padding: 1;`** --> No issues whatsoever
https://user-images.githubusercontent.com/12344434/173606603-b8be095e-41ab-4d22-9b88-fd8e3b163acf.mov

## Changelog

Added a prop `disableSafePadding?: boolean`. If `true`, it won't add the new style rule `safePadding` on the focus guide's view.

Not sure if its the best way, but right now there is no way to add some styling only to the focus guide view's style.

## Test Plan

```ts
import React, { FC, useRef, useState } from 'react';

import {
  FlatList,
  StyleSheet,
  Text,
  View,
  TouchableOpacity,
  findNodeHandle,
  TVFocusGuideView,
  ViewProps,
  ListRenderItem,
} from 'react-native';

type Node = ReturnType<typeof findNodeHandle>;

const swimlaneData = ['1', '2', '3'];

type TButtonProps = Pick<ViewProps, 'style'> & {
  text: string | number;
  onPress?: (id: Node) => void;
  pressedItemId: Node;
};

export const TButton: FC<TButtonProps> = ({ text, style, pressedItemId, onPress }) => {
  const [focused, setFocused] = useState(false);
  const ref = useRef(null);
  const pressed = pressedItemId === findNodeHandle(ref.current);
  return (
    <TouchableOpacity
      ref={ref}
      style={[style, { backgroundColor: pressed ? 'green' : focused ? 'red' : 'blue' }]}
      onFocus={() => {
        setFocused(true);
      }}
      onBlur={() => {
        setFocused(false);
      }}
      onPress={() => onPress?.(findNodeHandle(ref.current))}
      activeOpacity={0.7}
    >
      <Text style={{ color: 'white', fontSize: 18, fontWeight: 'bold' }}>{text}</Text>
    </TouchableOpacity>
  );
};

export const TestFocusChildless = () => {
  const [destinations, setDestinations] = useState<Node[]>([]);
  const [enabled, setEnabled] = useState(false);

[...]

PR #317, opened 2022-06-20
refactor: tvfocusguideview component and documentation

leonchabbey:
## Summary

Small refactor of the TVFocusGuideView following the merges of our PRs https://github.com/react-native-tvos/react-native-tvos/pull/313 and https://github.com/react-native-tvos/react-native-tvos/pull/314.

And added some documentation about TVFocusGuideView in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,11 +203,10 @@ class Game2048 extends React.Component {
 
 - _TVFocusGuideView_: This component provides support for Apple's `UIFocusGuide` API and is implemented in the same way for Android TV, to help ensure that focusable controls can be navigated to, even if they are not directly in line with other controls.  An example is provided in `RNTester` that shows two different ways of using this component.
 
-| Prop | Value | Description |  |  |
-|---|---|---|---|---|
-| destinations | any[] | Array of `Component`s to register as destinations of the FocusGuideView |  |  |
-| safePadding | 'both' (default) \| 'vertical' \| 'horizontal' \| null | When the FocusGuide children are exactly the same size as the FocusGuide's container, the focus will almost certainly be given directly to the children without going through the FocusGuide. This prop make sure it doesn't happen by adding a padding of 1 in all directions.<br />"null" to disable it. |  |  |
-|  |  |  |  |  |                                                                       |   |   |
+| Prop | Value | Description | 
+|---|---|---|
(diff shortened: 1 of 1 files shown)


douglowder (changes requested):
Thanks VERY much for the documentation additions and Typescript types!

I would ask that your refactor be done only on `TVFocusGuideView.android.js`, and leave `TVFocusGuideView.js` as is.

leonchabbey:
May i ask why you want separate files split per platform for TVFocusGuideView?

The way I refactored it avoid a lot of code duplication, is a cleaner implementation overall (IMO 😃) and at the same time highlights the differences on how the `destinations` prop is handled on tvOS and Android TV.

For example, the file split you did added [a regression](https://github.com/react-native-tvos/react-native-tvos/blob/tvos-v0.68.2/Libraries/Components/AppleTV/TVFocusGuideView.android.js#L70) on Android, the padding is not overridden on the container, which would have been avoided with this refactor.

douglowder:
I thought I tested the Android implementation with the new RNTester example and verified that the padding worked as expected. It is actually not supposed to be applied to the container, but to the contents, so I am not convinced that there is a regression :)

The reason I split the files is that I intend in the near future to unify them using the same pattern as the Apple TV implementation, via `Commands.setDestinations(nativeDestinations)`, so that everything works better with the Fabric architecture. However, I agree that this should be possible with your unified implementation as well.

I'll do some testing of your changes and then go ahead and merge.

douglowder (approved):
Thanks VERY much for the documentation additions and Typescript types!

I would ask that your refactor be done only on `TVFocusGuideView.android.js`, and leave `TVFocusGuideView.js` as is.

leonchabbey:
Sorry, the "regression" I was talking about is that if someone passes padding through the style prop, it will be also applied to the container because we are not overriding the paddingTop, paddingBottom, ... like we do for margins. I fixed that with [this change](https://github.com/react-native-tvos/react-native-tvos/pull/314/files#diff-8a312b8d2f6629fa4245b2aa7af7d1dc779997f031dc858bb5a96031e5706cd5R88-R91) and I saw you didn't do the same for Android :)

And thanks for the explanation about the future changes you have in mind! I agree with you, the more the platform implementations are similar the better!

PR #327, opened 2022-08-17
Add CHANNEL_UP, CHANNEL_DOWN, and DIGIT (0-9) key event support to Android TV

havlasme:
## Summary

Add CHANNEL_UP, CHANNEL_DOWN, and DIGIT (0-9) key event support to Android TV

## Changelog

[Android] [Added] - Add CHANNEL_UP, CHANNEL_DOWN, and DIGIT (0-9) key event support to Android TV

## Test Plan

We develop application that utilizes aforementioned events, we've made a build against react-native fork with these changes and it was working as expected. These changes just add more button mappings, so I don't think it requires some extensive testing.


freeboub (approved):

douglowder:
I assume this will have no effect on text input views?

havlasme:
No it does not have effect on text input. You can subscribe the digits with TvEventHandler and they still work in text input.

douglowder (approved):

PR #354, opened 2022-09-15
Apple TV: BackHandler subscribers should be called from last to first.

leonchabbey:
## Summary

To match Android's behaviour, the BackHandler subscribers should be called from last to first.


leonchabbey:
My bad, it already does this actually. It was an edge case in our code which was re-creating back handlers when it shouldn't.

PR #359, opened 2022-09-22
Use react-native version to find the hermes packages in podspec

geroale:
Fixed an issue where podspec was looking for the wrong hermes version (if enabled). Insted of the actual react-native-tvos (x.x.x-X) version, it should using the react-native (x.x.x) one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "react-native-tvos",
   "version": "0.69.5-0",
+  "rn-version": "0.69.5",
   "bin": "./cli.js",
   "description": "A framework for building native apps using React",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


douglowder:
Thanks for this @geroale ! I'm going to leave this open for now, as I am actively working on getting Hermes up and running on Apple TV and the podspec may change quite a bit.

douglowder:
Ok @geroale I'm fixing this in a different way so that Hermes works for iOS (not tvOS yet), until the full Hermes implementation is done for TV. See https://github.com/react-native-tvos/react-native-tvos/commit/5b913fbca33195aac2df8cbc7d29232197f70812

PR #371, opened 2022-10-06
Feature/focus visible android v0.68.2

SebastienChauvin:
## Summary

Focus handling on Android to be more similar to tvOS: invisible views cannot be navigated to by the user but can still be focused using the hasTVPreferredFocus property

Debugger can be toggled : 
```

  const {FocusModule} = ReactNative.NativeModules;
  FocusModule.setDebugger(true);
  FocusModule.setLog(true);

```
## Changelog

[CATEGORY] [TYPE] - Message


douglowder:
I find it very interesting that `hasTVPreferredFocus` works on invisible views... is this really the behavior we want?

douglowder:
@SebastienChauvin once you are finished adding commits to this PR, please let me know, and I'll take a look.

SebastienChauvin:
Of course will do. Sorry I had not realised I was on the official repo, I thought I was on our fork.

ReactAndroid/src/main/java/com/facebook/react/modules/focus/FocusModule.java
```diff
@@ -0,0 +1,54 @@
...
+  @ReactMethod
+  public void setDebugger(boolean enabled) {
+    FocusModule.enabled = enabled;
+    LocalBroadcastManager.getInstance(getReactApplicationContext()).sendBroadcast(new Intent(ON_CHANGE));
+  }
+
+  @ReactMethod
+  public void setLog(boolean enabled) {
+    FocusModule.log = enabled;
+  }
```
leonchabbey, line 53: I think it would be nice to add an additional check that the app is not running in production mode for both of those functions at least.
To make sure even if the developer forgot to remove his `FocusModule.setDebugger(true)` that it won't be enabled in production.

douglowder:
@SebastienChauvin @leonchabbey I went ahead and merged #370 to the 0.69.5 branch and have been testing it out. However, given that you are still actively working on this feature for your application, and given that I am not sure the `RNTester` example is working as intended, I am not going to include it in the upcoming 0.69.6-0 release.  I will continue to keep track on your progress with this PR for 0.68.2.

ReactAndroid/src/main/java/com/facebook/react/modules/focus/FocusModule.java
SebastienChauvin: I can imagine situations where one would want to have a debug focus in a release build.
Developers have the possibility to do the right thing, I don't see the point of doing anything for them.

SebastienChauvin:
>
Yes that sounds like a good strategy. We will update you when ready for 0.68 and do more tests later on 0.69

SebastienChauvin:
This is not perfect, I guess we will come back later with a proper PR for Fabric.

PR #423, opened 2022-12-29
fix: make sure UIFocusGuide is only applied when necessary

### LINKED ISSUE #421: [FABRIC ONLY] Every View has UIFocusGuide attached to it
### Description

Every view element gets a `UIFocusGuide` no matter if they're actually a focus guide or not. This is clearly a bug. I'll create a PR with the fix. Creating this issue for documenting the problem.

Here's how `TVFocusGuideExample` look like:

Screenshot 2022-12-29 at 13 40 35

And here's the main screen of `RNTester` app:

Screenshot 2022-12-29 at 13 48 15

### Version
0.69.6

### Steps to reproduce

- Build the `RNTester` app with Fabric enabled
- Use XCode's `Debug View Hierarchy` tool to see it

### LINKED ISSUE #422: [FABRIC ONLY] New UIFocusGuide constraint keep getting added
### Description

UIFocusGuide's layout constraints keep growing after renders, layout changes, and fast refreshes. I have a fix ready, will create a PR. It is kind of related to #421 but not entirely.

This is after the initial render (You can see even here it is doubled):

Screenshot 2022-12-29 at 13 55 34

After just a couple of fast refreshes:

Screenshot 2022-12-29 at 13 56 25

### Version

0.69.6

### Steps to reproduce

- Build the `RNTester` app with Fabric enabled
- Pick an element with `TVFocusGuide` attached to it as a reference point
- Check current `TVFocusGuide` constraints using XCode's `Debug View Hierarchy` tool
- Trigger fast refresh a couple of times
- Check `TVFocusGuide` constraints again, you'll see it's growing

yusufyildirim:
This PR fixes #421 and fixes #422.
## Summary

As it is described in #421, every element was getting `UIFocusGuide` attached to it unnecessarily. It was due to the `UIFocusGuide` creation in `initWithFrame` method without any checks. I moved the `UIFocusGuide` creation to `addFocusDestinations` to fix the issue.

The second issue was constant `UIFocusGuide` constraints growth as it is described in #422. The reason was setting constraints in `updateLayoutMetrics` which gets called very often. It was leading to unnecessary constraint definitions. Moved that section to `addFocusDestinations` method as well to fix it.

Here's before (doesn't show the constraints in this SS but you can check #422):

Screenshot 2022-12-29 at 13 40 35

Here's after the fix (even after a couple fast refreshes):

Screenshot 2022-12-29 at 14 14 17

## Test Plan

- Build the `RNTester` app (Fabric enabled)
- Navigate to `TVFocusGuideExample`
- Check if only the element that uses `TVFocusGuide` on the JS side has `TVFocusGuide` attached to them on the native side.
- Pick an element with `TVFocusGuide` attached to it as a reference point
- Check the `TVFocusGuide` constraints
- Trigger fast refresh a couple of times
- See if constraints grow

### DIFF
--- a/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
+++ b/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
@@ -71,8 +71,6 @@ using namespace facebook::react;
     _tvParallaxProperties.pressMagnification = 1.0f;
     _tvParallaxProperties.pressDuration = 0.3f;
     _tvParallaxProperties.pressDelay = 0.0f;
-    self.focusGuide = [UIFocusGuide new];
-    [self addLayoutGuide:self.focusGuide];
 #else
     self.multipleTouchEnabled = YES;
 #endif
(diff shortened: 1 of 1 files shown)


douglowder:
@yusufyildirim thanks VERY much for this patch. Looks good overall, I'll test it out and then merge if there are no issues.

douglowder (approved):

PR #425, opened 2022-12-30
fix: [Android TV] Text should not be accessible by default 

### LINKED ISSUE #389: Text getting focus instead of touchables (Android only)
### Description

I just upgraded to version `0.69.5-2` very meticulously and everything compiles and runs, but the focus engine seems to be focusing Text nodes now instead of Touchables. I discovered this by turning on a debug setting, then you can see the Text gets the focus instead of the button. With debug mode off it looks like the focus is not working until you click a few times (as we don't have focus states for text).

This was not the case in the previous version we were running (0.66.3-3).

### Version

0.69.5-2

### Output of `npx react-native info`

```
System:
    OS: macOS 12.6
    CPU: (8) arm64 Apple M2
    Memory: 122.89 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.1 - ~/.nvm/versions/node/v16.18.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.18.1/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.18.1/bin/npm
    Watchman: 2022.10.31.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /Users/wouterds/.rbenv/shims/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.1, iOS 16.1, macOS 13.0, tvOS 16.1, watchOS 9.1
    Android SDK:
      API Levels: 26, 28, 30, 31
      Build Tools: 28.0.3, 30.0.2, 30.0.3, 31.0.0
      System Images: android-31 | Google TV ARM 64 v8a
      Android NDK: Not Found
  IDEs:
    Android Studio: 2021.2 AI-212.5712.43.2112.8609683
    Xcode: 14.1/14B47b - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.17 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.0.0 => 18.0.0 
    react-native: Not Found
    react-native-macos: Not Found
    react-native-tvos:  0.69.5-2 
  npmGlobalPackages:
    *react-native*: Not Found
```

### Steps to reproduce

Render a screen with some touchables on top, then some text in between and then some more buttons. When navigating down from the top you will first pass through the text nodes and only then land on the touchables on the bottom.

### Snack, code example, screenshot, or link to a repository

I currently don't have a demo project to show this behaviour but can look into one if desired and we don't get this sorted out else.

douglowder:
## Summary

Recent changes in the `Text` component make it accessible and pressable by default. On Android TV, this has the side effect of making it focusable. Only controls (`Touchable`, `Pressable`, `TextInput`, etc.) should be focusable.

Fixed by changing `Text.js` so that a `Text` component must explicitly have `accessible={true}` in order to be pressable or focusable on TV platforms.

Fixes #389 .

## Test Plan

Tested with the `TVInputDemo` test app, and `RNTester`.

### DIFF
--- a/Libraries/Text/Text.js
+++ b/Libraries/Text/Text.js
@@ -15,6 +15,7 @@ import processColor from '../StyleSheet/processColor';
 import TextAncestor from './TextAncestor';
 import {NativeText, NativeVirtualText} from './TextNativeComponent';
 import {type TextProps} from './TextProps';
+import Platform from '../Utilities/Platform';
 import * as React from 'react';
 import {useContext, useMemo, useState} from 'react';
 
@@ -185,7 +186,7 @@ const Text: React.AbstractComponent<
         {...restProps}
         {...eventHandlersForText}
         disabled={_disabled}
-        accessible={accessible !== false}
+        accessible={Platform.isTV ? accessible === true : accessible !== false}
         accessibilityState={_accessibilityState}
         allowFontScaling={allowFontScaling !== false}
         ellipsizeMode={ellipsizeMode ?? 'tail'}
--- a/Libraries/Text/Text.js
+++ b/Libraries/Text/Text.js
@@ -186,7 +186,11 @@ const Text: React.AbstractComponent<
         {...restProps}
         {...eventHandlersForText}
         disabled={_disabled}
-        accessible={Platform.isTV ? accessible === true : accessible !== false}
+        accessible={
+          Platform.isTV && Platform.OS === 'android'
+            ? accessible === true
+            : accessible !== false
+        }
         accessibilityState={_accessibilityState}
         allowFontScaling={allowFontScaling !== false}
         ellipsizeMode={ellipsizeMode ?? 'tail'}


yusufyildirim:
This is an interesting topic. I don't think accessible elements should be focusable but there's a line that explicitly does that: https://github.com/react-native-tvos/react-native-tvos/blob/tvos-v0.69.6/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java#L58

Gotta check if it's really a necessity on Android but seemed strange to me 😄 I'm interested, and will check this one.

Your fix would do the trick as well, though. Thanks!

wouterds (approved):
Looks like a decent fix to me! Good that we can close this annoyance, will verify if it fixes some of the other issues that I experienced as well.

Happy new year doug!

douglowder:
Thanks @yusufyildirim ! One thing I do need to check is that this change doesn't break VoiceOver (which is supported on Apple TV at least). If that turns out not to be an issue, then we are good. If it is, then we need a different fix.

douglowder:
@yusufyildirim I'm making one more adjustment, to make this change happen only on Android TV. It is not needed on Apple TV and could affect VoiceOver there.

PR #426, opened 2022-12-30
feat: make TVFocusGuide work reliably without safePadding

yusufyildirim:
## Summary
Whenever I saw `safePaddings` approach, I thought "okay, this is a good workaround but the focus engine should be deterministic and shouldn't require this sort of workaround". Started digging to get rid of `safePaddings` necessity and found that it is fairly simple and the focus engine indeed works as expected if we prepare the layout in a way that it can do its job.

Here's how it works without the changes in this PR:

<table>
  <tr>
    <th>OLD ARCH</th>
    <th>FABRIC</th>
  </tr>
  <tr>
    <td>
    
https://user-images.githubusercontent.com/22980987/210111757-98a19aa3-5c33-4b8e-84ee-a85b4edcddd7.mov
</td>
        <td>
        
https://user-images.githubusercontent.com/22980987/210110060-80a31215-9fbe-4514-a85e-e0d6cc75ea1c.mov
</td>
  </tr>
</table>

This is `TVFocusGuideSafePaddingExample`. The expected behavior is simple. Button 2, 3, 4, and 5 are wrapped with `TVFocusGuide` for every row which redirects the focus to `Button 3`. If the focus moves toward that container, it should always be favored before its children (Button 2, 3, 4, 5) and redirect the focus to `Button 3`.

On my tests I couldn't find any problem with the old arch, it works perfectly fine already. But, if you check the Fabric screen recording, you'll see that it doesn't work as expected when `safePadding` is not provided.

I'd -intuitively- always expect tvOS focus engine (or any other focus engine, to be honest) to favor the container view before its children **if the focus is coming outside the container**. The problem is exactly this. With Fabric, iOS has `view flattening` built-in which flattens the UI tree as much as possible. So, *the children* are not really children in the native UI tree and in fact, because they are rendered later, the children views most likely gains priority over the "parent" container.

Here's how UI trees look like:

| OLD ARCH | FABRIC |
|---|---|
| Screenshot 2022-12-30 at 23 05 19 | Screenshot 2022-12-30 at 19 47 12 |

Everything is nested on the old arch since there's no view flattening but it's the complete opposite on Fabric and the tree is *very* "flat" 😄 tvOS can't figure out the right thing to do in this case.

## Solution
I set `collapsable` prop to false to prevent view flattening for `TVFocusGuide` view components and keep the parent-child relationship intact to fix the problem.

Here's how UI tree look like with the patch:

Screenshot 2022-12-30 at 20 54 49

And here's how it works with the patch:

https://user-images.githubusercontent.com/22980987/210115404-dadf38c7-62f6-4647-817b-6bec576b708b.mov

Here's how Focus Engine sees the views with and without the patch:
| W/O PATCH | WITH PATCH |
|---|---|
| context_8 | context_10 |

I did not remove any code and documentation related to `safePaddings` work but I think this PR makes it obsolete if I'm not missing any case. If we decide to merge this, we should almost completely revert these two PRs:

https://github.com/react-native-tvos/react-native-tvos/pull/314
https://github.com/react-native-tvos/react-native-tvos/pull/317

### DIFF
--- a/Libraries/Components/TV/TVFocusGuideView.js
+++ b/Libraries/Components/TV/TVFocusGuideView.js
@@ -97,7 +97,9 @@ const FocusGuideViewTVOS = (props: TVFocusGuideViewProps) => {
       Commands.setDestinations(hostComponentRef, nativeDestinations);
   }, [props.destinations]);
 
-  return <ReactNative.View ref={focusGuideRef} {...props} />;
+  return (
+    <ReactNative.View ref={focusGuideRef} {...props} collapsable={false} />
+  );
 };
 
 const FocusGuideViewAndroidTV = (props: TVFocusGuideViewProps) => {
--- a/Libraries/Components/TV/TVFocusGuideView.js
+++ b/Libraries/Components/TV/TVFocusGuideView.js
@@ -23,7 +23,7 @@ type TVFocusGuideViewProps = $ReadOnly<{
    * Useful for absolute focus guides without children
    * Defaults to true
    */
-  enabled: ?Boolean,
+  enabled?: boolean,
 
   /**
    * The views the focus should go to
@@ -31,7 +31,7 @@ type TVFocusGuideViewProps = $ReadOnly<{
   destinations: ?(Object[]),
 
   /**
-   * How the TVFocusGuideView content safe padding should be applied. "null" to disable it.
+   * @deprecated Don't use it, no longer necessary.
    */
   safePadding?: 'vertical' | 'horizontal' | 'both' | null,
 }>;
@@ -39,48 +39,20 @@ type TVFocusGuideViewProps = $ReadOnly<{
 const TVFocusGuideView = ({
   enabled = true,
   safePadding,
-  style,
-  ...otherProps
+  ...props
 }: TVFocusGuideViewProps): React.Node => {
-  const paddingChoice = safePadding === undefined ? 'both' : safePadding;
-  const focusGuidePadding =
-    paddingChoice === 'both'
-      ? {padding: 1}
-      : paddingChoice === 'vertical'
-      ? {paddingVertical: 1}
-      : paddingChoice === 'horizontal'
-      ? {paddingHorizontal: 1}
-      : null;
-
   const enabledStyle = {display: enabled ? 'flex' : 'none'};
+  const style = [styles.container, props.style, enabledStyle];
 
-  /**
-   * The client specified layout(using 'style' prop) should be applied the container view ReactNative.View.
-   * And the focusGuide's layout should be overridden to wrap it fully inside the container view.
-   * For example, if the client specifies 'marginLeft' property in the style prop,
-   * then the TVFocusGuideView will apply the 'marginLeft' for both the parentView and the focusGuideView.
-   * and so, the left margin is getting added twice and UI becomes incorrect.
-   * The same is applicable for other layout properties.
-   */
-  const focusGuideStyle = [focusGuidePadding, style, styles.focusGuide];
+  if (Platform.isTVOS) {
+    return <FocusGuideViewTVOS {...props} style={style} />;
+  }
 
-  return (
-    /**
-     * The container needs to have at least a width of 1 and a height of 1.
-     * Also, being the container, it shouldn't be possible to give it some padding.
-     */
-    <ReactNative.View style={[style, styles.container, enabledStyle]}>
-      {Platform.isTV ? (
-        Platform.isTVOS ? (
-          <FocusGuideViewTVOS {...otherProps} style={focusGuideStyle} />
-        ) : (
-          <FocusGuideViewAndroidTV {...otherProps} style={focusGuideStyle} />
-        )
-      ) : (
-        otherProps.children
-      )}
-    </ReactNative.View>
-  );
+  if (Platform.isTV) {
+    return <FocusGuideViewAndroidTV {...props} style={style} />;
+  }
+
+  return <ReactNative.View {...props} style={style} />;
 };
 
 const FocusGuideViewTVOS = (props: TVFocusGuideViewProps) => {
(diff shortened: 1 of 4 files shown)


yusufyildirim:
@leonchabbey @SebastienChauvin I'd appreciate it if you guys check that it works for you as well

douglowder:
- I have invited you, @leonchabbey, and @SebastienChauvin to be collaborators on this repo.
- I see that `collapsable` is marked Android-only in the RN docs: https://reactnative.dev/docs/view#collapsable-android , is that not really true?
- Please fix the merge conflict
- Again, thanks very much for your detailed discussions!

yusufyildirim:
@douglowder thank you very much! Yes, I think the documentation is outdated. View flattening used to be an android-only feature and that's why `collapsable` used to be *android only*. But the new renderer (Fabric) has a lot written in C++ and shares to logic between both platforms.

I think this is where the logic lies: https://github.com/react-native-tvos/react-native-tvos/blob/tvos-v0.69.6/ReactCommon/react/renderer/components/view/ViewShadowNode.cpp#L66

leonchabbey:
@yusufyildirim Thank you very much for this detailed PR and the research you did!
It's really good news that we still have a way to disable the view flattening on both platforms with the `collapsable` prop. We are procrastinating our migration to Fabric for a while because our custom Android focus management (hopefully a PR coming soon) depends entirely on the parent-child relation.

Unfortunately, I can't seem to get rn-tester running anymore on my machine... I get this error:
```
// React-Codegen / FBReactNativeSpec
No viable conversion from returned value of type 'std::optional<double>' to function return type 'folly::Optional<double>'

Xcode: 14.2
tvOS: 16.1
Fabric: both enabled or disabled
```
Already tried all the cleaning of Pods, build, ... but nothing fixed it.
Did any of you encountered this issue before?

About the PR changes, we still need to support the `safePadding` prop for the projects that are not using Fabric yet. Could you update the documentation accordingly and only apply the safePadding's style when fabric is disabled?

yusufyildirim:
@leonchabbey I've been following the focus work you guys are doing in your repo! I have an alternative solution that requires way less code and custom logic compared to your solution, guys. I'll create a PR and @ you to discuss it soon.

That aside, `safePadding` works as expected on non-Fabric tvOS projects already, that was one of my points. Is Android the real concern here? If is that so, the PR I'll be creating soon (as I mentioned above) should solve that problem.

Regarding the error, I've never faced this :/

leonchabbey:
@yusufyildirim Ok sorry I read too fast the PR description. So the "quick fix" provided by the safePadding is still necessary for both Fabric and non-Fabric right? The only difference is that `collapsable` is necessary for Fabric to disable the view flattening.

About the safePadding, it's actually necessary **only** for tvOS. The style is still applied on Android to keep the UI the same on both platforms (since it adds a few pixels to fix focus issues on tvOS).

Sure, tag us on your next PR when it's ready. We will have a look!

leonchabbey:
I confirm I was able to test the PR on 0.69.5 and it works as said.
Also, removing safePadding for both Fabric and non-Fabric doesn't seem to change anything. It still works when `collapsable={false}`.

I'd like to test again safePadding on 0.68.2 to be sure I didn't forget an edge case for which safePadding was necessary. Maybe something fixed this issue somehow with 0.69.x and safePadding is not useful anymore.
Hopefully I'll be able to do those tests tomorrow, now my machine has issues with building in 0.68.x...

yusufyildirim:
@leonchabbey Thanks for checking! Yes, It should work without `safePaddings` on both Fabric and old renderer. That's why I think we can and should remove it completely. Even if there's an edge case, I'm positive we can figure something out without the `safePadding` workaround. I'd be more than happy to fix it if you find any. Thanks again for spending your time on testing!

douglowder:
@leonchabbey for building and testing RNTester, I found that in some cases you need to execute

```
git clean -xfd
```

and then do yarn install and everything else from scratch. Hopefully that puts the repo back in a good state.

leonchabbey:
Thanks @douglowder! `git clean -xfd` fixed the project. Good to know! Could have saved me an hour yesterday 😄 

So yes, 0.68.x definitely still need safePadding. I was able to reproduce the bug from the examples. Something done in 0.69.x seems to have fixed this.
@yusufyildirim Since this PR will be merged in 0.69+ only, I think you can safely delete the code related to safePadding in this PR as well.

yusufyildirim:
@leonchabbey Great! Maybe it was related to one of these PRs: https://github.com/react-native-tvos/react-native-tvos/pull/420, https://github.com/react-native-tvos/react-native-tvos/pull/423 or https://github.com/react-native-tvos/react-native-tvos/pull/427. I'll do more extensive testing and remove everything related to `safePaddings`.

Removing the prop shouldn't be a breaking change but which approach would be the best here @douglowder? We can let it stay as a "dummy" prop that does nothing or remove it completely 😄

douglowder:
@yusufyildirim Let's leave it as a dummy prop and mark it deprecated in the comment block above (line 34 of `TVFocusGuidView.js`)

Libraries/Components/TV/TVFocusGuideView.js
```diff
@@ -23,64 +23,36 @@ type TVFocusGuideViewProps = $ReadOnly<{
    * Useful for absolute focus guides without children
    * Defaults to true
    */
-  enabled: ?Boolean,
+  enabled?: boolean,
```
yusufyildirim, line 26: This is a fix for a flow type error introduced in [here](https://github.com/react-native-tvos/react-native-tvos/pull/379/files#diff-d3b78a646e3e2034cd8994da0316617f6b0e4015bb1c2bb328930947dc6015e5R26). Not related to this PR but wanted to fix it while I'm in there 😄

yusufyildirim:
I specifically left `TVFocusGuideSafePaddingExample` because it's still useful for testing. I'll probably create a more advanced/real-world testing playground for the focus stuff in one of my upcoming PRs

douglowder (approved):

PR #447, opened 2023-01-11
fix: improve TVFocusGuide stability on Android

### LINKED ISSUE #439: [ANDROID] TVFocusGuide doesn't work as expected
### Description

Even though TVFocusGuide works ok horizontally, it is quite funky when we change the focus vertically. Here's the example:

https://user-images.githubusercontent.com/22980987/211202509-eb672594-9a65-4b09-8d9a-621877cb82e0.mov

All of Button 2, 3, 4, and 5 are wrapped with a `TVFocusGuide`. The focus should move to `Button 3`s on any vertical focus change action because `Button 3`s are the active `destinations`.

Here's how it works on tvOS:
[Will add this one later]

### Version
0.69.6

### Steps to reproduce

- Launch RNTester app
- Navigate to `TVFocusGuide examples with safePadding prop` example
- Focus to any `Button 2`
- Press up or down depending on which `Button 2` you are focused on
- Confirm the focus doesn't move to `Button 3`

yusufyildirim:
This PR fixes #439.

## Summary

As a starter, we need a high level of understanding of how Android's Focus engine works and how my changes will make an impact on it.

### How Android's Focus Engine works?

When you press a D-pad button, a `focusSearch` tree-traversal starts from the currently focused element all the way up to the root element. Every element *can* implement its own custom `focusSearch` logic and interrupt this. Once the root element is hit, it runs `findNextFocus` method from the `FocusFinder` instance. `FocusFinder` is *the* focus engine of Android.

`findNextFocus` checks if there's any user-specified focus (simply set by nextFocus* props) and immediately moves focus to the appropriate one if there's any. If there's no user-specified focus, it traverses every single descendant (from top to bottom this time) [to build a "focusable elements list"](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/FocusFinder.java;l=105?q=findnextfocus&ss=android%2Fplatform%2Fsuperproject) by calling their `addFocusables` method. Elements implement this `addFocusables` method to determine if they and/or their children should be accounted as focusable or not. Pretty much every element class extends `View` or `ViewGroup` class in Android. So, every element will inherit the default behavior from them. Here you can see how Android's `ViewGroup` [overrides `addFocusables`](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/ViewGroup.java;l=1309?q=addfocusables%20&ss=android%2Fplatform%2Fsuperproject). There you can even see how the [`descendantFocusability`](https://developer.android.com/reference/android/view/ViewGroup#attr_android:descendantFocusability) property of ViewGroup is [implemented](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/ViewGroup.java;drc=61d06bbec93e335119066679a8b2ed138883ab0c;l=1312?q=addfocusables%20&ss=android%2Fplatform%2Fsuperproject).

The important point is, there's no parent-child relationship for `FocusFinder`. It finds every focusable element, builds the array, and starts doing math calculations to find the best match for the next focusable element. It also doesn't know if the elements are off-screen or blocked by another element etc. It sounds quite inefficient and error-prone and it actually is IMHO.

Here's an example:

Screenshot 2023-01-11 at 21 41 46

There are 20 focusable elements on the screen. Here's the length of the list `FocusFinder` builds when we trigger a focus search:

Screenshot 2023-01-11 at 15 32 14

So it indeed checks for a potential match for every single element on the screen.

### What this PR change?
The problem with `FocusFinder`s approach is basically not caring about parent-child relationship. We exactly need that for `TVFocusGuide` to work reliably. We can't change the implementation but we can make it work as the way we want it.

We want none of the children of `TVFocusGuide` to be focusable *until* `TVFocusGuide` catches the focus and redirects it to one of its children. After that, children can attract focus, until focus leaves the `TVFocusGuide` borders again.

This PR exactly does this by overriding `addFocusables` method of `ReactViewGroup`. It checks if the focus is inside the `TVFocusGuide`. Marks the child elements as focusable if that's the case. If not, only marks the `TVFocusGuide` element itself as focusable so any child elements remain invisible to `FocusFinder`. It may be hard to understand/visualize in your mind so here's how `FocusFinder` sees the world after this changes:

When the focus is outside `TVFocusGuide`s, they hide their elements from `FocusFinder` and this is roughly sees:

Screenshot 2023-01-11 at 22 17 26

When we press right and move the focus to the `TVFocusGuide` on the right side it redirect the focus to its defined `destination` and `FocusFinder` starts to see the app like this:

Screenshot 2023-01-11 at 22 22 57

I hope the visualization is enough to understand it! Depending on how you structure your app using `TVFocusGuide`, it can even potentially improve the focus performance since there'll be much fewer candidates for the next focus. 

Here's the before and after:

<table>
  <tr>
    <th>BEFORE</th>
    <th>AFTER</th>
  </tr>
  <tr>
    <td>
    
https://user-images.githubusercontent.com/22980987/211202509-eb672594-9a65-4b09-8d9a-621877cb82e0.mov
</td>
        <td>
        

https://user-images.githubusercontent.com/22980987/211924115-69b51e2c-c857-45ad-b14e-818a32e2685a.mov
</td>
  </tr>
</table>

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java
@@ -60,6 +60,8 @@ import com.facebook.react.uimanager.common.UIManagerType;
 import com.facebook.react.uimanager.common.ViewUtil;
 import com.facebook.yoga.YogaConstants;
 
+import java.util.ArrayList;
+
 /**
  * Backing for a React View. Has support for borders, but since borders aren't common, lazy
  * initializes most of the storage needed for them.
@@ -1016,6 +1018,72 @@ public class ReactViewGroup extends ViewGroup
     }
     return false;
   }
+
+  private boolean isTVFocusGuide() {
+    return focusDestinations.length > 0;
+  }
+
+  @Nullable
+  private ReactViewGroup findParentFocusGuide(View view) {
+    ViewParent parent = view.getParent();
+
+    while (parent != null) {
+      if (parent instanceof ReactViewGroup) {
+        ReactViewGroup elem = (ReactViewGroup) parent;
+        if (elem.isTVFocusGuide()) return elem;
+      }
+      parent = parent.getParent();
+    }
+
+    return null;
+  }
+
+  /***
+   * This is meant to be used only for TVFocusGuide.
+   * @return View | null
+   */
+  @Nullable
+  private View getFocusedChildOfFocusGuide() {
+    if (!isTVFocusGuide()) return null;
+
+    /*
+     * We can have nested `TVFocusGuide`s, this is a typical scenario.
+     * The problem is, returned element from `getFocusedChild` can be
+     * either a direct `child` of the `TVFocusGuide` or a `descendant` of it.
+     * Let's say if we run `getFocusedChild` for the Root element, it will always
+     * give us an element even though the Root element is not the direct parent
+     * of the focused element.
+     * So, we need to find the closest `TVFocusGuide` to the focused child
+     * to make sure the focused element's closest `TVFocusGuide` is "this" one.
+     */
+    View focusedChild = this.getFocusedChild();
+    if (focusedChild == null) return null;
+
+    ReactViewGroup parentFocusGuide = findParentFocusGuide(focusedChild);
+    if (parentFocusGuide == this) return focusedChild;
+
+    return null;
+  }
+
+  @Override
+  public void addFocusables(ArrayList<View> views, int direction, int focusableMode) {
+    if (isTVFocusGuide()) {
+      View focusedChild = getFocusedChildOfFocusGuide();
+
+      /*
+       * We only include the view's (TVFocusGuide) itself if it doesn't have a child that currently has the focus.
+       * Otherwise, it means focus is already in the `TVFocusGuide` and all of it's descendants should be included
+       * for focus calculations.
+       */
+      if (focusedChild == null) {
+        views.add(this);
+        return;
+      }
+    }
+
+    super.addFocusables(views, direction, focusableMode);
+  }
+
 
   @Override
   public boolean requestFocus(int direction, Rect previouslyFocusedRect) {


wouterds:
Looks good to me, but I don't know too much about Java/native Android side so going to leave approval over to @douglowder

leonchabbey:
Looks good to me as well!
I'll ask @SebastienChauvin to have a look since he worked quite a bit with Android's focus system for [our app](https://github.com/react-native-tvos/react-native-tvos/compare/tvos-v0.68.2...SRGSSR:react-native-tvos:latest/srgssr).

douglowder (approved):
This looks great. I will see if I can come up with a test screen that has nested focus guides, and see if everything works as expected on both Apple TV and Android TV.

SebastienChauvin:
I agree, this seems to be a good way to make focus guides work with the focus finder.

In our project, we still _had_ to work around the focus finder because the focus Search has other "funky" behaviours not connected to tv guides. The concept of beam does not work for us: we don't want to navigate to something that is far but perfectly aligned, we'd rather navigate to a view that is just in the direction.

But integrating this would not hurt anyway and might simplify/fix some of the focusSearch workarounds we had.

yusufyildirim:
@SebastienChauvin @leonchabbey Thanks for putting your time into this! Yeah, Android's search has very weird behaviors 😄 I'm thinking that we can eliminate most of the problems by constructing our layouts by leveraging `TVFocusGuide`, though. There are still a couple of fixes/feature additions to go to get there. Right after those, I'll write a blog post or create a discussion here about "how we should construct our layout to eliminate the focus problems" and we can discuss.

SebastienChauvin:
Coming out of all the struggle we had with the FocusFinder, I was pretty convinced we would need to build our own focus engine to make things consistent and understandable by the developer.
I am curious to see what you can find. Looking at the craziness in the FocusFinder code (`beamBeats` in particular), I don't see how we could achieve something clean, consistent that still goes through FocusFinder for every focus search. I don't mind being proven wrong thought !

A very different approach we are considering in the team is to use a custom FocusFinder algorithm for Android, something similar to what is implemented here: https://github.com/NoriginMedia/Norigin-Spatial-Navigation for instance. This looks simple and works well for the ReactJS port of our app.

yusufyildirim:
@SebastienChauvin Totally feel you. I've been there. Thought that it is a must to write a custom Focus engine for Android that works just like Norigin's Spatial Navigation or tvOS's focus engine 😄 Your work inspired me, been following your rn tvos fork and the custom focus logic you guys are working on. That is totally viable but I decided to keep it as my nuclear option because a different, simpler idea appeared in my head 😄

We can't fix Android's FocusFinder but we might make it work the way we want it by guiding it heavily with very little effort on the user (our) side. Maybe my idea won't scale, we'll see!

wouterds:
>_I think the goal with the focus on this repo should be to make it work the same as much as possible between the two platforms, ideally the developer won't have to do change anything based on the platform._

YES! 💯

There are some inconsistencies right now or APIs that behave differently (I've talked about this before in some of the issues here), but ideally they don't. That's the whole point of RN, right? Write once.

yusufyildirim:
I would say React Native's vision is to have parity on both platforms *as much as possible* with respect to the platform defaults. When it comes to a topic like this, changing the native platform's behavior completely should be avoided in the core repo if you ask me. If we need a unified focus solution, it should be a community effort *ideally*. Building anything custom comes with a maintenance and documentation/education burden.

For example, we don't change any default platform behavior with `TVFocusGuide` here. It's an addition. If you don't use it, everything should work how they supposed to work depending on your platform. This is our attempt to transform `TVFocusGuide` into a cross-platform primitive that will help you tackle your focus problems without changing the platform.

douglowder:
@yusufyildirim this fix looks good -- I've tested it in RNTester and in TVInputDemo on both Android TV emulator and device.

I enhanced the focus guide example in TVInputDemo to include nested focus guides and the behavior seems correct and stable. (https://github.com/douglowder/TVInputDemo/commit/6b8bf8e6d057e54123724addb746b1f576b32a98)

I have not tested it in Fabric yet.

Feel free to merge once you are ready.

PR #450, opened 2023-01-19
feat: add auto focus management feature to TVFocusGuide

yusufyildirim:
## Motivation

Focus management is known to be one of the hardest and most annoying parts of TV application development. It is the fundamental part of TV UX that we have to make it work right and intuitively. Creating such an experience requires two things; one is **predictable focus navigation** and two, **automatic focus redirection**. Native components support *automatic focus redirection* part, like [UICollectionView](https://developer.apple.com/documentation/uikit/uicollectionview/1618022-rememberslastfocusedindexpath), and some of the Android components (even though I couldn't find anything documented). But still, they may fall short on the *predictability* part, especially Android. Nevertheless, developers don't have access to those native components in RN due to some valid reasons, and the *focus navigation predictability* is totally a different problem that requires attention,  I felt a need to build a solution. That's why we have the `autoFocus` feature for `TVFocusGuide` today.

## Solution
This PR introduces a new prop for `TVFocusGuide` called `autoFocus` which helps developers worry less about two things, one is **automatic focus redirection** and the second is **focus recovery**. Let's take a look at what I mean by those:

### Automatic Focus Redirection

It's very common in TV apps to expect the focus to move to the very first element of the container focus moves when jumping between focus containers (regions, collections, or whatever you'd like to call them) instead of the closest element on the screen to the currently focused element. Let's visualize this to make it easy to understand:

Simulator Screen Shot - Apple TV - 2023-01-26 at 17 10 22

I press `right` while the focus is on the very bottom element of the `Side Menu` container and expect the focus to move to the very first element of the content container on the right, which is the one I marked in green. Not the one I marked in red even though it is literally the one on the right side of the currently focused element. The reason is simple, we almost always want users to start experiencing the content from the beginning, not from a random place.

This is only true for the first visits, though. We want users to continue from where he/she left off on the subsequent visits. Here's the expectation:

https://user-images.githubusercontent.com/22980987/214903694-9d05d9a0-924e-46d8-8c63-26691b40159e.mov

`autoFocus` feature redirects the focus to the very first element on the first visit but also remembers the last focused element and redirects the focus back to it on subsequent visits which is the element at the very bottom on "Side Menu" container case and "Category-1" on the content container case. 

I think this example alone makes it easy to understand the benefit and how developers can leverage it but what about the predictability part?

### How it helps with predictability?

Let me show you a tricky case and how both platforms handle it natively, without `TVFocusGuide` and `autoFocus` feature:

<table>
  <tr>
    <th>Android</th>
    <th>tvOS</th>
  </tr>
  <tr>
    <td>
    
https://user-images.githubusercontent.com/22980987/214909820-4a9c8d78-5b20-43bc-9c5d-4aee398c4384.mov
</td>
        <td>
        

https://user-images.githubusercontent.com/22980987/214909768-f1a9c0a2-e3ed-4649-8e41-693f8dba513e.mov
</td>
  </tr>
</table>

Two completely different behavior. tvOS is even kinda random, which is weird. Both of the platforms have their own, fundamentally different approaches to decide where the focus should go next, so differences are expected but, not a good experience for TV devs. I'll describe more on that in the **"How to utilize TVFocusGuide?"** section.  But now, let's see how it works on both platforms with `autoFocus` feature enabled:

 <table>
  <tr>
    <th>Android</th>
    <th>tvOS</th>
  </tr>
  <tr>
    <td>
    
https://user-images.githubusercontent.com/22980987/214914389-fa13d9e0-3a0e-4290-937d-2e196e4904e5.mov
</td>
        <td>
        

https://user-images.githubusercontent.com/22980987/214914464-7215efa9-e019-4362-af18-bd8281c29342.mov
</td>
  </tr>
</table>

Voilà! By wrapping the focusable elements with distinct purposes (like, side menu, content container, every category, and category item list, etc.) by `TVFocusGuide` we restrict the focus navigation paths and force the focus to go to the first or last focused element of the container. Since this behavior is guaranteed cross-platform, we end up with predictable focus navigation behavior. I'll describe how we achieve that in the "How to utilize TVFocusGuide?" section.

### Focus Recovery
`TVFocusGuide` *tries to detect* if the currently focused element gets removed from the screen and *tries to redirect* the focus to the very first focusable element in it. I'm saying *tries* because it works quite well in a lot of cases but may not be able to successfully recovery in every case that involves scrollable/virtualized content due to some constraints. It is due to how ScrollView/FlatList works, not a `TVFocusGuide` problem, I'll take a look at that later.

Let's see what it does:

https://user-images.githubusercontent.com/22980987/214921678-eb83f306-2d2f-4c2a-9c5f-114b07834c8a.mov

Whenever we `press` on one of the `Restore Focus Test` list items, it toggles a state that forces the list to render with brand new items. So, the currently focused element gets removed from the screen. When that happens, the closest `TVFocusGuide` reclaims the focus for a brief moment and waits for the UI thread to finish its job (drawing the new elements) and then redirects the focus to the very first element. That's it. It's also how it works on tvOS, the implementation is different, though. The system manages it quite well with very minimal guidance.

Still, this feature may not sound that necessary because why would you remove the focused element from the screen, right? Let's see the case that forced me to build this mechanism. In other words, what happens **without** the focus recovery mechanism:

https://user-images.githubusercontent.com/22980987/214925507-533d5a74-5c52-444b-86cb-b17036421d1f.mov

While `Category-1` item is focused, I press `right` and `down` arrows super fast. Moving the focus `right` triggers a re-render on the `item list` below and `down` makes the focus to move to the first element on the list. The problem is, this happens super fast while React preparing the list to re-render, the `Category 1 - Item-0` item gains the focus and immediately gets removed from the screen because React finishes rendering and informs the native side to paint the new elements. Oh, and Android has the behavior that moves the focus to the item closest to the top-left corner when the currently focused item gets removed. That's why the focus moves there.

## How to utilize TVFocusGuide?

We should first understand how tvOS's and Android's focus engine works. There's one simple example it'll make everything clear:

 <table>
  <tr>
    <th>Android</th>
    <th>tvOS</th>
  </tr>
  <tr>
    <td>
    
https://user-images.githubusercontent.com/22980987/214931270-3a5c3ad8-f13a-4bee-ad78-498adb2421fe.mov
</td>
        <td>
        

https://user-images.githubusercontent.com/22980987/214931202-39719a3e-2c2a-447c-99a8-c07ce054391d.mov
</td>
  </tr>
</table>

tvOS doesn't do anything when we press `left` because it searches for an element on the left that *precisely aligns* with the currently focused element. Android on the other side, searches for *anything close* to the currently focused element on the left and there we have an item in the side menu which gains the focus. That is the fundamental difference. 

Android would favor too if there was a perfectly aligned item, just like tvOS looking for. So they both would work the same in that case.

**So, we can't change how these two engines work but we can design our layouts in a way that there will always be one obvious next focus candidate, for both engines! That's what we should always aim for, having only one possible focus candidate in every direction (top, right, bottom, left).**

### How to achieve that?
You must leave no gap without `TVFocusGuide`. It doesn't mean you should put it everywhere, though. Use it sparingly, only where it is really necessary. You should construct your layout in a way that focus can jump to only one candidate. 

As I [described in my previous PR](https://github.com/react-native-tvos/react-native-tvos/pull/447), `TVFocusGuide` component hides its descendant focusable elements from the platform focus engine until it receives and redirects the focus to one of its children.  This is a mechanism to make sure `TVFocusGuide`s always receives focus before its descendants. If you place `TVFocusGuide`s in a grid-like form, there will always be one and only one candidate --which is the next focus guide in the direction you wanna move -- since they hide their descendant focusables. This is the fundamental idea.

Here's how the new TVFocusGuide playground is constructed:

Screenshot_1674767864

`TVFocusGuide` borders are highlighted in random colors to make them evident. Pink arrows show the directions focus can flow through. There is always one and only one possibility and a solid one. Everything is so aligned that it leaves zero questions for focus engines and they do their job just fine. Simplifying the equation for them 😄 

I wrote a *crude* TVFocusGuide debugger to make it even easier to explain how underlying focus engines see the focusable elements on UI when you wrap them with `TVFocusGuide`:

https://user-images.githubusercontent.com/22980987/214963375-10ce45fd-c9fb-47f8-bf8b-e4526581eebf.mov

(Don't mind the flickers in this video, I wrote this thing in 10 minutes to demonstrate my point. There's a constant unnecessary show/hide action going on 🤣)

There's an extra important part I wanna point out. Pay attention Genres / Sub Genres / Year columns. They look like breaking our one simple rule. They are three separate focus containers in the way of our vertical focus flow. But, if you pay attention, you'll see there's another `TVFocusGuide` wrapping them to make them fit in to the definition.

https://user-images.githubusercontent.com/22980987/214966469-457234ec-f41c-4a5f-9db0-0b49285b82af.mov

That's all folks. I tried my best to explain every important thing as visual as possible. You deserve a medal if you made it this far.

Long story short:

76qnb7


douglowder:
Thanks for this, it looks really good! I will set this up and test it locally and then add review comments. 🎉🎉

yusufyildirim:
@SebastienChauvin @leonchabbey @wouterds I told you guys before that [I'll write a blog post](https://github.com/react-native-tvos/react-native-tvos/pull/447#issuecomment-1381389337) and then we can discuss. Scratch that. Consider this as the blog post I promised 😄 Looking forward to hearing your feedback!

wouterds:
I'm gonna have a proper look this evening, sounds really promising. Thanks for all the effort @yusufyildirim!

packages/rn-tester/js/examples/TVFocusGuide/TVFocusGuideAutoFocusExample.js
```diff
@@ -0,0 +1,323 @@
...
+
+const getItemText = ({item, prefix}) => {
+  return prefix ? `${prefix}-${item}` : `${item}`;
+};
+
+const HList = ({
+  itemCount,
+  itemWidth = 500 * scale,
+  itemHeight = 120 * scale,
+  itemColor = FOCUSABLE_ITEM_COLOR,
```
douglowder, line 136: Same comment -- we should use a theme color

packages/rn-tester/js/examples/TVFocusGuide/TVFocusGuideAutoFocusExample.js
```diff
@@ -0,0 +1,323 @@
...
+    </TVFocusGuide>
+  );
+};
+
+const RestoreFocusTestList = () => {
+  const [randomize, setRandomize] = React.useState(false);
+  const data = React.useMemo(() => generateData(10, randomize), [randomize]);
+  /**
+   * This is a test to make sure that the focus is restored
+   * after the list is re-rendered and currently focused item is removed.
```
douglowder, line 225: I tested this and it works quite well on both platforms.

I did notice that when the list is rerendered, the focus goes to the leftmost item currently on screen, not the first item in the list -- this is probably what we want anyway as it preserves horizontal scroll position.

packages/rn-tester/js/examples/TVFocusGuide/TVFocusGuideAutoFocusExample.js
```diff
@@ -0,0 +1,323 @@
...
+    alignItems: 'center',
+  },
+  rowTitle: {marginRight: 10 * scale, fontSize: 24 * scale},
+  container: {
+    flex: 1,
+    flexDirection: 'row',
+  },
+  sideMenuContainer: {
+    width: 100 * scale,
+    backgroundColor: 'gray',
```
douglowder, line 305: Probably want a theme color here too

douglowder (approved):
Looks good, just found a few things in the example code 👍

packages/rn-tester/js/examples/TVFocusGuide/TVFocusGuideAutoFocusExample.js
```diff
@@ -0,0 +1,323 @@
...
+    <RNText style={[styles.text, {color: theme.LabelColor}, style]}>
+      {children}
+    </RNText>
+  );
+};
+
+const FocusableBox = ({
+  width,
+  height,
+  color: backgroundColor = FOCUSABLE_ITEM_COLOR,
```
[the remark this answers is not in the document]
douglowder: Done

packages/rn-tester/js/examples/TVFocusGuide/TVFocusGuideAutoFocusExample.js
douglowder: Done

packages/rn-tester/js/examples/TVFocusGuide/TVFocusGuideAutoFocusExample.js
douglowder: Done

packages/rn-tester/js/examples/TVFocusGuide/TVFocusGuideAutoFocusExample.js
yusufyildirim: Yeah, this is known and expected. The scroll state should be reset first if you want it to move the focus to the *very* first item of the list. Scrollable content (especially virtualized) can lead to interesting behavior as I pointed out in the description. It's a separate problem that I'll take a look at later, though. Just wanted to leave a comment here for future readers.

wouterds:
When I compiled `android` with these changes, dropped it into my project, and tried to compile it it fails because of a conflict related to `react-native-screens` it seems like.

```> Task :react-native-screens:compileReleaseKotlin FAILED
e: /Users/wouterds/Projects/telenet/multibrand/app/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt: (22, 5): Accidental override: The following declarations have the same JVM signature (setAutoFocus(Z)V)```

yusufyildirim:
> When I compiled `android` with these changes, dropped it into my project, and tried to compile it it fails because of a conflict related to `react-native-screens` it seems like.
> 
> 
> 
> ```
> 
> > Task :react-native-screens:compileReleaseKotlin FAILED
> 
> e: /Users/wouterds/Projects/telenet/multibrand/app/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt: (22, 5): Accidental override: The following declarations have the same JVM signature (setAutoFocus(Z)V):
> 
>     fun `<set-autoFocus>`(`<set-?>`: Boolean): Unit defined in com.swmansion.rnscreens.SearchBarView
> 
>     fun setAutoFocus(p0: Boolean): Unit defined in com.swmansion.rnscreens.SearchBarView
> 
> ```

Very interesting finding. Thank you! Will take a look

yusufyildirim:
@wouterds @douglowder Ok, so they have this `SearchBarView` which extends `ReactViewGroup`. I introduced a public method in `ReactViewGroup` called `setAutoFocus` and they have [this line](https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt#L22) which generates an invisible `setAutoFocus` method I believe (Kotlin magic). I'll create a new PR to rename that public method to something else or completely get rid of it to prevent any potential conflict. Thanks for reporting!

matias-sandell:
This is probably my favorite PR of all time. Thanks for some stellar autofocus work @yusufyildirim! 🥇

PR #465, opened 2023-02-16
feat: add focusTrap functionality to TVFocusGuide

yusufyildirim:
## Summary

This PR introduces a new addition to `TVFocusGuide` arsenal, `focusTrap*`. This new functionality allows us to *trap* the focus inside a container. When it's in, we can make sure it stays in until we don't want it to do so.

This concept is well-known and used on the web for a11y. The difference here is though, our implementation allows us to have fine-grained control based on the direction. We can trap the focus in only one direction while allowing it to move freely in other directions. E.g we can trap it for **left** and **right** actions but not for **up** and **down**.

This flexibility is important because the functionality is mainly implemented to fix a focus bug that exists in **VirtualizedList** family (FlatList etc.). I'm not even sure calling it a *bug* is even fair but let me explain it first.

## The "bug"

As a refresher, this piece of information is important; Focus engines find the next focusable based on the rendered UI elements. This principle conflicts with virtualization because its sole purpose is *not* to render the elements until they are about to be seen by the user. In theory, the rendering speed of the virtualized list will be faster than the user's scroll speed so the user should never notice this optimization. In reality, though, we end up seeing good ol' blank areas.

When rendering speed can't keep up with user's navigation speed, the list runs out of items, and the focus moves to the most logical next element on the next d-pad command which is possibly outside the list. This is a horrible UX and it happens much more than you think it'd happen.

image

It's easier to explain by showing:

https://user-images.githubusercontent.com/22980987/219511510-4ac10e15-ee89-4cf7-803b-8bd77b21e2cf.mov

I created a list with artificially slow components. Each component takes at least 200ms to render --which is a bit extreme but I've seen worse in actual prod apps, lol--. Also, locked `windowSize` to 1 so it doesn't render items in advance. You can see the whole config [here](https://github.com/react-native-tvos/react-native-tvos/pull/465/files#diff-74d1b0c793b4986766077b9926b5e0a12ccac8f518ea4ee4d85561a22f515671R269).

As I spam `right` on the d-pad, we reach the **end**(!) of the list and focus moves to the focusable item `RIGHT`. If you pay attention to the end of the list, you'll see that the next items get rendered late and focus escapes outside. 

## Solution
Implemented `trapFocus*` props for `TVFocusGuide` that helps us control the focus routes by direction. Used it to fix VirtualizedList's focus *bug* by simply wrapping the children of the list and enabling/disabling `trapFocus*` props based on the list's direction and `first`/`last` rendered item information. If `last` is less than provided `data` array's `length`, it means the very last item is not rendered, yet there's still a risk of the bug's occurrence. We keep trapping the focus in that case. As simple as this. You can figure out the rest by reading the code I hope.

With this in place, the focus stays inside until reaching the very beginning or the very end of the list even if the user spams d-pad actions.

### tvOS

<table>
  <tr>
    <th>tvOS (Before)</th>
    <th>tvOS (After)</th>
  </tr>
  <tr>
    <td>
    
https://user-images.githubusercontent.com/22980987/219516155-a33c797d-1d6e-424b-bbe1-0028e0fd0ea3.mov
</td>
        <td>
        
https://user-images.githubusercontent.com/22980987/219517775-5029ec7e-9b80-45a4-b72f-21645d465b06.mov
</td>
  </tr>
</table>

### Android

<table>
  <tr>
    <th>Android (Before)</th>
    <th>Android (After)</th>
  </tr>
  <tr>
    <td>
    
https://user-images.githubusercontent.com/22980987/219516010-ec5a4038-5ac1-4c1d-a391-5c1015f3e045.mov
</td>
        <td>
        

https://user-images.githubusercontent.com/22980987/219516022-d859b66d-38e5-40f2-a7b5-5ca56f8e0972.mov
</td>
  </tr>
</table>

### DIFF
--- a/React/Views/RCTTVView.m
+++ b/React/Views/RCTTVView.m
@@ -482,20 +482,4 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder : unused)
   self.focusGuide.preferredFocusEnvironments = destinations;
 }
 
-- (void)setTrapFocusUp:(BOOL)enabled {
-  self->_trapFocusUp = enabled;
-}
-
-- (void)setTrapFocusDown:(BOOL)enabled {
-  self->_trapFocusDown = enabled;
-}
-
-- (void)setTrapFocusLeft:(BOOL)enabled {
-  self->_trapFocusLeft = enabled;
-}
-
-- (void)setTrapFocusRight:(BOOL)enabled {
-  self->_trapFocusRight = enabled;
-}
-
 @end
--- a/React/Views/RCTTVView.m
+++ b/React/Views/RCTTVView.m
@@ -255,6 +255,23 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder : unused)
   motionEffectsAdded = NO;
 }
 
+- (BOOL)shouldUpdateFocusInContext:(UIFocusUpdateContext *)context
+{
+  // This is  the `trapFocus*` logic that prevents the focus updates if
+  // focus should be trapped and `nextFocusedItem` is not a child FocusEnv.
+  if ((_trapFocusUp && context.focusHeading == UIFocusHeadingUp)
+     || (_trapFocusDown && context.focusHeading == UIFocusHeadingDown)
+     || (_trapFocusLeft && context.focusHeading == UIFocusHeadingLeft)
+     || (_trapFocusRight && context.focusHeading == UIFocusHeadingRight)) {
+
+    // Checks if `nextFocusedItem` is a child `FocusEnvironment`.
+    // If not, it returns false thus it keeps the focus inside.
+    return [UIFocusSystem environment:self containsEnvironment:context.nextFocusedItem];
+  }
+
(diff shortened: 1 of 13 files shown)


React/Views/RCTTVView.m
```diff
@@ -465,4 +482,20 @@ - (void)addFocusGuide:(NSArray*)destinations {
   self.focusGuide.preferredFocusEnvironments = destinations;
 }
 
+- (void)setTrapFocusUp:(BOOL)enabled {
+  self->_trapFocusUp = enabled;
```
douglowder, line 486: Not sure you really need these setters, these are what get synthesized by the "assign" property declarations.

douglowder (approved):
Looks good! Had one question about tvOS implementation, see inline.

React/Views/RCTTVView.m
yusufyildirim: Yeah, you're absolutely right. They're totally unnecessary 😄 Removed ✅

PR #469, opened 2023-03-06
feat: refactor TVFocusGuide implementations and introduce `setDestinations` function

yusufyildirim:
## Summary
This PR makes `autoFocus` and `destinations` features work together in harmony, opening up many possibilities. `destinations` prop takes precedence when used with `autoFocus` together. I'll give 
a practical example down below. Also, it introduces `setDestinations` imperative method which can be used to programmatically set destinations outside React render cycle which can be useful.

To be able to achieve these, I refactored the TVFocusGuide component and its native implementations a bit, but not entirely. Removed Android and iOS distinctions inside `TVFocusGuide` component code by unifying `destinations` management part. We use native method calls for both platforms now.

## How it works?
As I said earlier, we can use now `autoFocus` and `destinations` together. `autoFocus` does its job in the background, redirecting the focus to the first focusable element in the container and keeping the last focused element in memory. The thing is, if `destinations` property is provided (or, called through ref.setDestinations) it takes precedence and instead of moving focus to the first focusable or last focused item (if there is any), focus moves to one of the `destinations`.

This combinations makes it possible [implementing such a feature](https://github.com/react-native-tvos/react-native-tvos/pull/469/files#diff-74d1b0c793b4986766077b9926b5e0a12ccac8f518ea4ee4d85561a22f515671R253-R292) which forces the focus to move to a specific element on the first visit but than delegates the rest to the `autoFocus` functionality.

<table>
  <tr>
    <th>tvOS</th>
    <th>Android</th>
  </tr>
  <tr>
    <td>
    

https://user-images.githubusercontent.com/22980987/223598364-1218793b-029b-4487-a2b5-ea1b2a04c98b.mov
</td>
        <td>
        

https://user-images.githubusercontent.com/22980987/223598379-e7d424d7-68d1-4bae-94d1-6b109ef30ead.mov
</td>
  </tr>
</table>


React/Views/RCTTVView.m
```diff
@@ -482,4 +484,43 @@ - (void)addFocusGuide:(NSArray*)destinations {
   self.focusGuide.preferredFocusEnvironments = destinations;
 }
 
+- (void)removeFocusGuide {
+	if (self.focusGuide != nil) {
+        focusDestinations = nil;
+        previouslyFocusedItem = nil;
+
+		[self removeLayoutGuide:self.focusGuide];
```
douglowder, line 492: nit: fix indentation here

packages/rn-tester/android/app/build.gradle
```diff
@@ -95,7 +95,7 @@ react {
 
 project.ext.react = [
     enableCodegen: true, // Keep this here until it's sync'ed to Android template.
-    enableFabric: (System.getenv('USE_FABRIC') ?: 'true').toBoolean()
+    enableFabric: (System.getenv('USE_FABRIC') ?: 'false').toBoolean()
```
douglowder, line 98: Generally RNTester assumes Fabric is true these days (this is also the case for iOS/tvOS, see the Podfile), so I think this should be reverted.

tvos-types.d.ts
```diff
@@ -104,15 +106,20 @@ declare module 'react-native' {
...
+
   /**
    * This component provides support for Apple's `UIFocusGuide` API,
    * to help ensure that focusable controls can be navigated to,
-   * even if they are not directly in line with other controls.
-   * An example is provided in `RNTester` that shows two different ways of using this component.
-   * https://github.com/react-native-tvos/react-native-tvos/blob/tvos-v0.63.4/RNTester/js/examples/TVFocusGuide/TVFocusGuideExample.js
+   * even if they are not directly in line with other controls on both tvOS and Android.
+   * An example is provided in `RNTester` that shows different ways of using this component.
+   * https://github.com/react-native-tvos/react-native-tvos/blob/tvos-v0.69.6/packages/rn-tester/js/examples/TVFocusGuide/TVFocusGuideAutoFocusExample.js
```
douglowder, line 120: We may change the name of the branch to 0.69.8 since we just merged in upstream changes from the core 0.69 release branch :)

douglowder (approved):
Looks great as usual! See a few comments and requests inline.

I'll also test this locally to see how things work on RNTester and TVInputDemo 😃

packages/rn-tester/android/app/build.gradle
yusufyildirim: Lol, I should stop creating PRs at 3 a.m.😂

douglowder: Obviously that's the best time to push code 😄

PR #477, opened 2023-03-17
fix: ts type declarations

### LINKED ISSUE #476: Build error with Unexpected token when importing react-native/tvos-types.d.ts
### Discussed in https://github.com/react-native-tvos/react-native-tvos/discussions/474

<div type='discussions-op-text'>

<sup>Originally posted by **matias-sandell** March 16, 2023</sup>
Trying out the new https://www.npmjs.com/package/react-native-tvos/v/0.69.8-0rc3 release and getting a build error when a TypeScript component imports `react-native/tvos-types.d`.

```
error SyntaxError: ......../node_modules/react-native/tvos-types.d.ts: Unexpected token (115:21)

  113 |   export type FocusGuideMethods = {
  114 |     setDestinations: (
> 115 |       destinations: (?React.ElementRef<HostComponent<unknown>>)[],
      |                      ^
  116 |     ) => void;
  117 |   }
  118 |.

FAILURE: Build failed with an exception.
```

Screenshot 2023-03-16 at 11 16 03

Workarounds
- Removing the `react-native/tvos-types.d` import from the .tsx file makes the build work again (and the app works when building for production, but not when running it through the react native dev server)
- Removing the `?` from the type definition also makes the app work through the dev server.

It's probably something with my setup but any hints as to what the underlying problem might be will be greatly appreciated! 🙌

The setDestinations type definition was introduced in https://github.com/react-native-tvos/react-native-tvos/commit/7143d7877ec432337f36693fba75f611178a96e8#diff-e14263f50370e82b314d0b19b9a21e70e0a0948798e28872703fdd1093b3610cR111</div>

yusufyildirim:
Fixes #476.

## Summary
Fixes a syntax error inside tvOS typescript declaration file. Also enriches `TVFocusGuideView` type.

### DIFF
--- a/tvos-types.d.ts
+++ b/tvos-types.d.ts
@@ -112,7 +112,7 @@ declare module 'react-native' {
 
   export type FocusGuideMethods = {
     setDestinations: (
-      destinations: (?React.ElementRef<HostComponent<unknown>>)[],
+      destinations: (React.ElementRef<HostComponent<unknown>> | null)[],
     ) => void;
   }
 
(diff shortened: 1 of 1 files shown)


tvos-types.d.ts
```diff
@@ -112,7 +112,7 @@ declare module 'react-native' {
 
   export type FocusGuideMethods = {
     setDestinations: (
-      destinations: (?React.ElementRef<HostComponent<unknown>>)[],
+      destinations: (React.ElementRef<HostComponent<unknown>> | null)[],
```
yusufyildirim, line 115: Nullability is pretty much a convenience thing. React refs are nullable by default and users will be passing refs as destinations. We let destinations nullable and filter out null values under the hood so devs won't have to place null checks everywhere

douglowder:
```suggestion
      destinations: (React.ElementRef<HostComponent<unknown>> | null | undefined)[],
```

douglowder (approved):

PR #485, opened 2023-04-06
fix: make sure initialScrollIndex is bigger than 0

okwasniewski:
## Summary

When passing `initialScrollIndex` smaller than 0 for example -1 new elements won't render. `_updateCellsToRender` function was checking if `this.props.initialScrollIndex` is truthy and -1 was returning true.

## Changelog

[General] [Fixed] - make sure initialScrollIndex is bigger than 0


okwasniewski:
@douglowder thanks, for taking a look! Yes, I've created similar PR for upstream. Should I decline this one?

douglowder:
Let me check what happens upstream... if it gets merged then I'll cherry pick from there

okwasniewski:
Hey, @douglowder the PR got merged to RN core.

PR #486, opened 2023-04-07
fix: restore focus responsibly

yusufyildirim:
## Summary
The focus recovery feature that is introduced in https://github.com/react-native-tvos/react-native-tvos/pull/450 had a bug. `hasTVPreferredFocus` or `requestTVFocus` commands were getting somewhat ignored during the focus recovery process. The implications were different on Android and tvOS. Set up a new test case to demonstrate the problem.

## Android

On Android, we detect and start a recovery effort when the focus is getting lost. Upon detection, the focus gets "parked" to the closest `TVFocusGuide` and we postpone the recovery process to the end of the main thread to give UI some time to complete rendering before we decide where the focus should go. The problem is, the focus can move to another item during that small timeframe. Like, one of the elements on the UI can request focus.  We weren't checking that case, this PR fixes that.

<table>
  <tr>
    <th>Android (Before)</th>
    <th>Android (After)</th>
  </tr>
  <tr>
    <td>
    

https://user-images.githubusercontent.com/22980987/230523857-386ffa3f-0891-4b67-a621-bd50d8074993.mov
</td>
        <td>
        

https://user-images.githubusercontent.com/22980987/230523870-98bf001c-8205-4afa-a94f-dfa93db5496f.mov
</td>
  </tr>
</table>

## tvOS

On tvOS however, the story is a bit different. The focus recovery logic works totally differently compared to Android. The main problem here wasn't the recovery itself but the way we move focus to the requested element. It was async by default which was leading `hasTVPreferred` prop or `requestTVFocus` actions (focusSelf) to be delayed to the end of the main queue. So, the focus was getting recovered to the initial element since there's no better candidate, and than getting recovered to the actual element afterward. This behavior was leading to a "stutter" as can be clearly seen in the screen recording. Introduced `requestFocusSelf` function which tries to move the focus synchronously by default but fallback to async if necessary.

<table>
  <tr>
    <th>tvOS (Before)</th>
    <th>tvOS (After)</th>
  </tr>
  <tr>
    <td>
    

https://user-images.githubusercontent.com/22980987/230523798-ab6e1303-7bbb-4a18-b985-8166d115a80b.mov
</td>
        <td>
        

https://user-images.githubusercontent.com/22980987/230523834-41feb460-c7fd-43fd-b8e3-481d6525a9f8.mov
</td>
  </tr>
</table>

### DIFF
--- a/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
+++ b/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
@@ -138,6 +138,36 @@ using namespace facebook::react;
   return (RCTRootComponentView *)rootview;
 }
 
+/// Handles self-focusing logic. Shouldn't be used directly, use `requestFocusSelf` method instead.
+-(bool)focusSelf {
+  RCTRootComponentView *rootview = [self containingRootView];
+  if (rootview == nil) return false;
+  
+  if (self.focusGuide != nil) {
+    rootview.reactPreferredFocusEnvironments = self.focusGuide.preferredFocusEnvironments;
+  } else {
+    rootview.reactPreferredFocusedView = self;
+  }
+
+  [rootview setNeedsFocusUpdate];
+  [rootview updateFocusIfNeeded];
+  return true;
+}
+
+/// Tries to move focus to `self`. Does that synchronously if possible, fallbacks to async if it fails.
+-(void)requestFocusSelf {
+  bool focusedSync = [self focusSelf];
+  
+  if (!focusedSync) {
+    // `focusSelf` function relies on `rootView` which may not be present on the first render.
+    // `focusSelf` fails and returns `false` in that case. We try re-executing the same action
+    // by putting it to the main queue to make sure it runs after UI creation is completed.
+    dispatch_async(dispatch_get_main_queue(), ^{
+      [self focusSelf];
+    });
+  }
+}
+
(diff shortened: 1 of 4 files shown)


douglowder (approved):

matias-sandell:
Nice work! Any chance we could get a 0.69.8-2 release with this fix @douglowder?

yusufyildirim:
@matias-sandell Yes, @douglowder will test and prep a new release whenever he finds some time

douglowder:
@matias-sandell we have a 0.69.8-2rc0 release candidate on npm if you want to help test.

matias-sandell:
Thanks, will try it today!

matias-sandell:
Did some testing and this fix didn't solve my particular focus issue (the `TVFocusGuideView` wrapping the menu items seem to get focus when doing a `navigation.navigate()` to a react-navigation screen the user has previously visited even though i set it to an item to the right of the menu using `requestTVFocus()`) but it also doesn't seem to break anything for me.

I'll continue using my workaround for the time being. (Which is to render a `View` with un-focusable menu items instead of the `TVFocusGuideView` around the menu items until everything else on the screen has rendered).

yusufyildirim:
@matias-sandell Interesting, yeah definitely doesn't sound related to this PR. I can take a look if you can provide *a very basic* reproduction

matias-sandell:
Thanks, that's very kind of you! It might take a while but I'll definitely try to create a simple test case demonstrating the issue.

PR #498, opened 2023-05-08
TVTouchable: Skip if findNodeHandle thrown error

jhen0409:
## Summary

I got the following error with component used TVTouchable on unmount.

`ExceptionsManager.js:158 Error: Unable to find node on an unmounted component.`

## Changelog

[General] [Fixed] - Skip if findNodeHandle thrown error in TVTouchable

## Test Plan

I'm upgrading react-native-tvos to 0.71 for a project, and it's working fine with this fix (by patch).

### DIFF
--- a/Libraries/Components/Touchable/TVTouchable.js
+++ b/Libraries/Components/Touchable/TVTouchable.js
@@ -36,7 +36,11 @@ export default class TVTouchable {
     this._tvEventHandler = new TVEventHandler();
     this._tvEventHandler.enable(component, (_, tvData) => {
       tvData.dispatchConfig = {};
-      if (ReactNative.findNodeHandle(component) === tvData.tag) {
+      let tag;
+      try {
+        tag = ReactNative.findNodeHandle(component);
+      } catch (e) {}
+      if (tag === tvData.tag) {
         if (tvData.eventType === 'focus') {
           config.onFocus(tvData);
         } else if (tvData.eventType === 'blur') {


Libraries/Components/Touchable/TVTouchable.js
```diff
@@ -36,7 +36,11 @@ export default class TVTouchable {
     this._tvEventHandler = new TVEventHandler();
     this._tvEventHandler.enable(component, (_, tvData) => {
       tvData.dispatchConfig = {};
-      if (ReactNative.findNodeHandle(component) === tvData.tag) {
+      let tag;
+      try {
+        tag = ReactNative.findNodeHandle(component);
+      } catch (e) {}
```
douglowder, line 42: I am not sure that this is the right fix for the issue.

The error that is thrown by `findNodeHandle()` in dev mode is important to know about, so we don't want to always suppress it.

Seems like there might be a deeper issue -- maybe unmounting is not cleaning up or disabling the handler correctly or in the right order.

douglowder (changes requested):
See inline comments.

Do you have a test app that reproduces the issue you found here?

Libraries/Components/Touchable/TVTouchable.js
douglowder: Does this problem occur with Fabric turned on, or with Fabric turned off?

jhen0409:
I've a minimum example for reproduce this issue.

```js
const [showFirst, setShowFirst] = useState(true);
return (
  <>
    {showFirst && (
      <TouchableOpacity onPress={() => setShowFirst(!showFirst)}>
        <Text>First</Text>
      </TouchableOpacity>
    )}
    {!showFirst &&
      [0, 1, 2].map(key => (
        <TouchableOpacity
          key={String(key)}
          onPress={() => setShowFirst(!showFirst)}
        >
          <Text>Key {key}</Text>
        </TouchableOpacity>
      ))}
  </>
);
```

The issue will be caused on `setShowFirst(!showFirst)` with unmounting the 3 touchables. It doesn't happen when its number is less than 3.

After try to reproduce that, I think the main problem is some event issue from TVEventHandler, although I haven't had time to figure out clearly, I'll open a new issue for this.

Libraries/Components/Touchable/TVTouchable.js
jhen0409: It's fabric turned off.

PR #499, opened 2023-05-09
fix: Support TV Typescript types properly in 0.71

douglowder:
## Summary

React Native 0.71 now provides its own Typescript types.  The TV-specific types previously in `tvos-types.d.ts` have now been moved to the `types` directory to be included automatically with all the other types.

The `tvos-types.d.ts` file is added back to the release distribution in order to not break any existing apps that try to import it. This file now just contains a comment pointing to the new location.

## Test Plan

Tested with a new template-created app.


wouterds (approved):
>The tvos-types.d.ts file is added back to the release distribution in order to not break any existing apps that try to import it. This file now just contains a comment pointing to the new location.

But users do no longer need to manually include the types then, right?

yusufyildirim (approved):
🚀

douglowder:
@wouterds that's correct, users no longer have to include a special types file for TV.

Issue #500, opened 2023-05-10
[TVTouchable] Error: Unable to find node on an unmounted component.


jhen0409:
### Description

I got the following error with component used TVTouchable on unmount.

`ExceptionsManager.js:158 Error: Unable to find node on an unmounted component.`

Platform: tvOS, Fabric is turned off.

### Version

0.71.7-0rc1

### Output of `npx react-native info`

```
info Fetching system and libraries information...
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M2
    Memory: 103.58 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.18.1 - ~/.nvm/versions/node/v16.18.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.18.1/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.18.1/bin/npm
    Watchman: 2023.03.20.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.4, iOS 16.4, macOS 13.3, tvOS 16.4, watchOS 9.4
    Android SDK: Not Found
  IDEs:
    Android Studio: 2021.2 AI-212.5712.43.2112.8815526
    Xcode: 14.3/14E222b - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.15 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.2.0 => 18.2.0 
    react-native: Not Found
    react-native-macos: Not Found
    react-native-tvos:  0.71.7-0rc1 
  npmGlobalPackages:
    *react-native*: Not Found
```

### Steps to reproduce

I've a minimum example for reproduce this issue, please see the code example below.

The issue will be caused on `setShowFirst(!showFirst)` with unmounting the 3 touchables. It doesn't happen when its number is less than 3.

\* From https://github.com/react-native-tvos/react-native-tvos/pull/498#issuecomment-1541157999.

### Snack, code example, screenshot, or link to a repository

```js
const [showFirst, setShowFirst] = useState(true);
return (
  <>
    {showFirst && (
      <TouchableOpacity onPress={() => setShowFirst(!showFirst)}>
        <Text>First</Text>
      </TouchableOpacity>
    )}
[...]

PR #502, opened 2023-05-10
fix: Remove TVTouchable interference between onPress and unrender

douglowder:
Fix issue #500 by adding a flag to the `TVTouchable` class to ensure that its tvEventHandler code is not called after destroy() is called.

## Test Plan

- Test code in #500 should work with no errors.

### DIFF
--- a/Libraries/Components/Touchable/TVTouchable.js
+++ b/Libraries/Components/Touchable/TVTouchable.js
@@ -43,11 +43,17 @@ export default class TVTouchable {
           config.onBlur(tvData);
         } else if (tvData.eventType === 'select') {
           if (!config.getDisabled()) {
-            config.onPress(tvData);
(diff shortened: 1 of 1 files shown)


jhen0409:
After tried this patch, I also found the same issue from listener of `BackHandler`. (`MENU` key)

douglowder:
@jhen0409 thanks for the additional testing. I also found the issue reproduces in Fabric. Will investigate further.

douglowder:
@jhen0409 I believe the second commit implements the correct fix.

douglowder:
@jhen0409 I tested this patch with your example code, and also with RNTester (that includes BackHandler), and no longer see any issues. I'm going ahead with merging this and create a new 0.71 release candidate for you and others to test with. Please let me know if there are any further issues.

PR #547, opened 2023-07-20
Smooth Scroll: Customize scroll duration and behavior for Android React native application: ScrollView, FlatList, & SectionList

PuttarajuPM:
Introduction
Enable the React Native Android applications to customize the scroll duration and behavior.

Details
Currently all scrollable views in React Native like ScrollView, FlatList, SectionList, etc uses Android OS specific hardcoded scroll duration and behavior.
Expose an API from ReactScrollViewHelper which gets called before applying an animation to the scroll.
API can pass both ScrollView and ValueAnimator which application can use to update the scroll duration and behavior.
This API will only be applicable when app uses ScrollView.scrollTo* OR FlatList.scrollTo* OR SectionList.scrollTo*

Example changes:
https://github.com/react-native-tvos/react-native-tvos/compare/tvos-v0.71.11...PuttarajuPM:react-native-tvos:v0.68.2-8
ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollViewHelper.java

Usage in Application:
https://github.com/PuttarajuPM/SmoothScrollAndroid/blob/main/android/app/src/main/java/com/smoothscrollandroid/MainApplication.java#L99

Current scroll behavior:
https://github.com/PuttarajuPM/SmoothScrollAndroid/blob/main/verified/current_scroll.webm

https://github.com/react-native-tvos/react-native-tvos/assets/91516440/9ca92bdf-9de5-41ee-9040-469255d09ec7

New scroll behavior:
https://github.com/PuttarajuPM/SmoothScrollAndroid/blob/main/verified/smooth_scroll_android.webm

https://github.com/react-native-tvos/react-native-tvos/assets/91516440/444e40dc-8139-4d21-bc46-70fb654e79b2

## Changelog:

[Android] [Changed] - Provide an API to customize the scroll speed and behavior on Android - only applicable when application requests manual scrollTo.


douglowder:
We will take a look at this. Two comments:

- Please remove all the changes and additions in the `android` directory, and also remove the version change in `gradle.properties`.
- This appears to be a change to code that is inherited from the React Native core repo, so it might be better to submit this PR to that repo.

douglowder (changes requested):
We will take a look at this. Two comments:

- Please remove all the changes and additions in the `android` directory, and also remove the version change in `gradle.properties`.
- This appears to be a change to code that is inherited from the React Native core repo, so it might be better to submit this PR to that repo.

PR #553, opened 2023-08-09
feat: add showsScrollIndex prop to ScrollView

okwasniewski:
## Summary

Add new prop to ScrollView that allows to hide the fast scrolling indicator for tvOS. Reference: https://developer.apple.com/documentation/uikit/uiscrollviewindexdisplaymode?language=objc

Before (default behaviour): 

https://github.com/react-native-tvos/react-native-tvos/assets/52801365/319898ed-b874-4822-a8bb-b4165e085130

After: 

https://github.com/react-native-tvos/react-native-tvos/assets/52801365/51e5fc82-e209-41ff-9d97-b2a4f027ad56

## Changelog

[tvOS] [Added] - new prop for controlling scrollIndex behaviour

## Test Plan

Add `showsScrollIndex` prop to flat list and check if the fast scrolling indicator shows up.

### DIFF
--- a/Libraries/Lists/VirtualizedList.d.ts
+++ b/Libraries/Lists/VirtualizedList.d.ts
@@ -344,4 +344,10 @@ export interface VirtualizedListWithoutRenderItemProps<ItemT>
   windowSize?: number | undefined;
 
   CellRendererComponent?: React.ComponentType<any> | undefined;
+
+  /**
+   * (TvOS only)
+   * Defines if UIScrollView index should be shown when fast scrolling.
+   */
+  showsScrollIndex?: boolean;
 }
--- a/Libraries/Lists/VirtualizedListProps.js
+++ b/Libraries/Lists/VirtualizedListProps.js
@@ -258,6 +258,11 @@ type OptionalProps = {|
    * The legacy implementation is no longer supported.
    */
   legacyImplementation?: empty,
+  /**
+   * (TvOS only)
+   * Defines if UIScrollView index should be shown when fast scrolling.
+   */
+  showsScrollIndex?: ?boolean,
 |};
 
 export type Props = {|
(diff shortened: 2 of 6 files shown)


okwasniewski:
Hey @douglowder can you take a look at this PR?

douglowder:
The PR looks fine! There are a couple of concerns...

- This native API seems to be for both iOS and tvOS. Have you submitted this to the React Native core repo?
- In 0.72 and later, the virtualized lists have been moved to a separate package within the new React Native monorepo structure. I'll have to think about how best to implement your proposed change in this repo, as I am attempting to make as few changes as possible in virtualized lists with the 0.72 release I'm working on now for TV.

douglowder:
@okwasniewski after review, it seems to me that these changes are great to have, but need to be submitted as a PR to the core repo, because of the refactor to `@react-native/virtualized-lists`.

NickGerleman:
> @okwasniewski after review, it seems to me that these changes are great to have, but need to be submitted as a PR to the core repo, because of the refactor to `@react-native/virtualized-lists`.

This change shouldn’t need to touch it. VirrualizedList already accepts and passes any Scrollview prop.

Out of tree platforms will want a story for being able to fork monorepo packages. But in this case I think it can be avoided.

I asked @okwasniewski to keep this change here, since the prop only does anything on tvOS, which is a confusing experience to have as part of the main RN types, documentation, etc.

NickGerleman:
An implementation note, this change is only for Paper. In RN core we have not been accepting any new Paper only APIs because they will go away when we switch folks over to Fabric. I’m not sure how tvos is handling this, but it would be most future proof to add the same change to the component view for Fabric.

douglowder:
@NickGerleman thanks for the comments! I'll reopen the PR for the TV repo and work with @okwasniewski to make sure it works for both Paper and Fabric.

douglowder:
Reopening. @okwasniewski please rebase this PR to the tvos-v0.72.4 branch.

okwasniewski:
@douglowder I couldn't reopen this PR so I've created a new one: https://github.com/react-native-tvos/react-native-tvos/pull/568 - I will work on Fabric implementation there

PR #568, opened 2023-08-30
feat: add showsScrollIndex prop to VirtualizedList

okwasniewski:
## Summary

Add new prop to ScrollView that allows to hide the fast scrolling indicator for tvOS. Reference: https://developer.apple.com/documentation/uikit/uiscrollviewindexdisplaymode?language=objc

Before (default behaviour): 

https://github.com/react-native-tvos/react-native-tvos/assets/52801365/319898ed-b874-4822-a8bb-b4165e085130

After: 

https://github.com/react-native-tvos/react-native-tvos/assets/52801365/51e5fc82-e209-41ff-9d97-b2a4f027ad56

## Changelog

[tvOS] [Added] - new prop for controlling scrollIndex behaviour

## Test Plan

Add `showsScrollIndex` prop to flat list and check if the fast scrolling indicator shows up.

### DIFF
--- a/packages/react-native/Libraries/Components/ScrollView/ScrollView.d.ts
+++ b/packages/react-native/Libraries/Components/ScrollView/ScrollView.d.ts
@@ -528,6 +528,13 @@ export interface ScrollViewPropsIOS {
    * The current scale of the scroll view content. The default value is 1.0.
    */
   zoomScale?: number | undefined;
+
+  /**
+   * (TvOS only)
+   * Defines if UIScrollView index should be shown when fast scrolling.
+   * Defaults to true.
+   */
+  showsScrollIndex?: boolean;
 }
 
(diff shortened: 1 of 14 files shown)


okwasniewski:
Hey @douglowder 

I've added implementation for Fabric. 

As I understood from what @NickGerleman wrote here: https://github.com/react-native-tvos/react-native-tvos/pull/553#issuecomment-1698335588

I can move the types from `VirtualizedList` to `ScrollView` as "VirtualizedList already accepts and passes any Scrollview prop" 

I've also set the default for this prop to be `true` so only if the user passes false explicitly it will be hidden.

douglowder (approved):

douglowder:
This looks fine now -- I'll merge and then test in RNTester, maybe add to one of the examples there.

PR #586, opened 2023-10-18
Fix/okwasniewski/text input focus

okwasniewski:
## Summary

Hey! 

This PR fixes issues with `TextInput` focus. Previously TV platform was using the implementation that was used for mobiles. 

The main difference is that now when `requestFocus()` method is called we actually set focus to the TextEdit component instead of returning whether component is focused. 

This PR also aligns commands that those views can receive, now TextInput can also receive `requestTVFocus` (a command that's used to focus normal views). As a side note I've also added `requestTVFocus` for the old arch for ViewGroup. 

Before: 

https://github.com/react-native-tvos/react-native-tvos/assets/52801365/c7225740-f1cd-4f76-8213-90bedbaf0ad1

After: 

https://github.com/react-native-tvos/react-native-tvos/assets/52801365/4980ada7-3973-4056-b7e0-a1e283fa8d88

## Changelog

[ANDROID] [FIXED] - TextInput is now focusable

## Test Plan

Launch RN tester example and check that you can navigate through TextInputs without issues.

### DIFF
--- a/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -368,13 +368,7 @@ public class ReactEditText extends AppCompatEditText
 
   @Override
   public boolean requestFocus(int direction, Rect previouslyFocusedRect) {
-    // This is a no-op so that when the OS calls requestFocus(), nothing will happen. ReactEditText
-    // is a controlled component, which means its focus is controlled by JS, with two exceptions:
-    // autofocus when it's attached to the window, and responding to accessibility events. In both
-    // of these cases, we call requestFocusInternal() directly.
-    // Always return true if we are already focused. This is used by android in certain places,
-    // such as text selection.
-    return isFocused();
+    return requestFocusInternal();
   }
 
   private boolean requestFocusInternal() {
--- a/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
+++ b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
@@ -147,6 +147,7 @@ public class ReactTextInputManager extends BaseViewManager<ReactEditText, Layout
   private static final int BLUR_TEXT_INPUT = 2;
   private static final int SET_MOST_RECENT_EVENT_COUNT = 3;
   private static final int SET_TEXT_AND_SELECTION = 4;
+  private static final int REQUEST_TV_FOCUS = 5;
 
   private static final int INPUT_TYPE_KEYBOARD_NUMBER_PAD = InputType.TYPE_CLASS_NUMBER;
   private static final int INPUT_TYPE_KEYBOARD_DECIMAL_PAD =
@@ -280,7 +281,7 @@ public class ReactTextInputManager extends BaseViewManager<ReactEditText, Layout
 
   @Override
   public @Nullable Map<String, Integer> getCommandsMap() {
-    return MapBuilder.of("focusTextInput", FOCUS_TEXT_INPUT, "blurTextInput", BLUR_TEXT_INPUT);
+    return MapBuilder.of("focusTextInput", FOCUS_TEXT_INPUT, "blurTextInput", BLUR_TEXT_INPUT, "requestTVFocus", REQUEST_TV_FOCUS);
   }
 
   @Override
@@ -288,6 +289,9 @@ public class ReactTextInputManager extends BaseViewManager<ReactEditText, Layout
       ReactEditText reactEditText, int commandId, @Nullable ReadableArray args) {
     switch (commandId) {
       case FOCUS_TEXT_INPUT:
+        this.receiveCommand(reactEditText, "focus", args);
+        break;
+      case REQUEST_TV_FOCUS:
         this.receiveCommand(reactEditText, "focus", args);
         break;
       case BLUR_TEXT_INPUT:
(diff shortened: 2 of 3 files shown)


douglowder (approved):
Looks good to me. Will merge after review by @yusufyildirim

packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
```diff
@@ -357,6 +358,9 @@ public void receiveCommand(ReactViewGroup root, int commandId, @Nullable Readabl
         handleSetDestinations(root, args);
         break;
       }
+      case CMD_REQUEST_TV_FOCUS:
+        root.requestFocus();
+        break;
```
yusufyildirim, line 363: I believe this is not related to new/old arch difference, to be accurate. There used to be `integer` commands and we were handling them the way you did it here but they're deprecated for a while. Commands are sent as string from JS side and there shouldn't be a way to send `integer` commands. The changes in this file should be safe to remove

Related PR: https://github.com/facebook/react-native/commit/3cae6fa950eed54bccccf61a28c0e85d5a004c6c#diff-71bc3f19ebb0dbc0c2f2c7c2fc24368d9bcf26118b43da76b1493a71f9375860R35

packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
```diff
@@ -368,13 +368,7 @@ public void clearFocus() {
...
   @Override
   public boolean requestFocus(int direction, Rect previouslyFocusedRect) {
-    // This is a no-op so that when the OS calls requestFocus(), nothing will happen. ReactEditText
-    // is a controlled component, which means its focus is controlled by JS, with two exceptions:
-    // autofocus when it's attached to the window, and responding to accessibility events. In both
-    // of these cases, we call requestFocusInternal() directly.
-    // Always return true if we are already focused. This is used by android in certain places,
-    // such as text selection.
-    return isFocused();
+    return requestFocusInternal();
```
yusufyildirim, line 371: I'm not super comfortable with this one. We should understand the decision made here better and build manual test cases to make sure everything works as expected before and after.

yusufyildirim (commented):
Hey, left some comments. I'm not super comfortable with the change. Even though it looks small, it's probably not. We should probably build some manual test cases in RNTester project first and do some PR archive digging to see why this decision was made in the core. Thanks for the contribution!

packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
```diff
@@ -290,6 +291,9 @@ public void receiveCommand(
       case FOCUS_TEXT_INPUT:
         this.receiveCommand(reactEditText, "focus", args);
         break;
+      case REQUEST_TV_FOCUS:
+        this.receiveCommand(reactEditText, "focus", args);
+        break;
```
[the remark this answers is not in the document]
okwasniewski: Mostly for consistency across views. So that we don't need to check if it's TextInput. 

Adding it on JS side won't support dispatchViewManagerCommand (that's what we are using internally to request focus). 

I think sticking to requestTVFocus across every component would be more predictable

packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
okwasniewski: Building additional test cases would be great. 

As far as I understand the reasoning behind this is that requestFocus was not used for mobile as there it makes sense to control focus from JS.

But for TVs currently you are unable to focus TextInput. I've been using a Touchable wrapper to onFocus call searchRef.focus() so this feels like a hack. (Maybe there is a better solution?)

If you take a look at requestFocusInternal it handles all cases like softKeyboardFocus etc. so I think we are covered there.

But I agree that this small looking change might be a breaking  change for some apps as this will change the behavior to work properly 😅

douglowder: @okwasniewski @yusufyildirim I think the right way to handle this change would be to put it behind a `ReactFeatureFlags` property (we did this for developers who needed to get key down events from the keyboard/DPAD).

satya1395:
looks like this i broken in latest version.

{
  "name": "TVTest",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "android": "react-native run-android",
    "ios": "react-native run-ios",
    "lint": "eslint .",
    "start": "react-native start",
    "test": "jest"
  },
  "dependencies": {
    "react": "18.3.1",
    "react-native": "npm:react-native-tvos@0.77.0-0"
  },
  "devDependencies": {
    "@babel/core": "^7.25.2",
    "@babel/preset-env": "^7.25.3",
    "@babel/runtime": "^7.25.0",
    "@react-native-community/cli": "15.0.1",
    "@react-native-community/cli-platform-android": "15.0.1",
    "@react-native-community/cli-platform-ios": "15.0.1",
    "@react-native/babel-preset": "0.77.0",
    "@react-native/eslint-config": "0.77.0",
    "@react-native/metro-config": "0.77.0",
    "@react-native/typescript-config": "0.77.0",
    "@types/jest": "^29.5.13",
    "@types/react": "^18.2.6",
    "@types/react-test-renderer": "^18.0.0",
    "eslint": "^8.19.0",
    "jest": "^29.6.3",
    "prettier": "2.8.8",
    "react-test-renderer": "18.3.1",
    "typescript": "5.0.4"
  },
  "engines": {
    "node": ">=18"
  }
}

PR #640, opened 2024-01-18
feat: README and example for TV-specific file extensions

douglowder:
## Summary:

Provide instructions and an example Metro config (in the template) for developers who need to support TV-specific file extensions (file.tv.tsx, etc.)

Closes #630 .

## Changelog:

[General][Added] Instructions for TV-specific file extensions

## Test Plan:

Tested in multiple local projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,19 @@ Besides most of the core Expo modules, these also work on TV:
 TV does NOT support dev client (dev menu, dev launcher) at this time.
 TV does NOT support Expo Router at this time.
 
+## _(New)_ How to support TV specific file extensions
+
+The template contains an [example Metro configuration](./packages/react-native/template/metro.config.js) that allows Metro to resolve application source files with TV-specific code, indicated by specific file extensions (e.g. `*.ios.tv.tsx`, `*.android.tv.tsx`, `*.tv.tsx`).
+
+When this is enabled, Metro will resolve files in the following order of preference:
+
+- `file.ios.tv.tsx` or `file.android.tv.tsx`
+- `file.tv.tsx`
+- `file.ios.tsx` or `file.android.tsx`
+- `file.tsx`
+
+This config is not enabled by default, since it will impact bundling performance, but is available for developers who need this capability.
+
 ## Code changes
 
 - _JavaScript layer_: Support for TV has been added to the `Platform` React Native API.
--- a/README.md
+++ b/README.md
@@ -90,6 +90,8 @@ Besides most of the core Expo modules, these also work on TV:
 TV does NOT support dev client (dev menu, dev launcher) at this time.
 TV does NOT support Expo Router at this time.
 
+See also the [Building Expo apps for TV](https://docs.expo.dev/guides/building-for-tv/) guide from Expo.
+
 ## _(New)_ How to support TV specific file extensions
 
 The template contains an [example Metro configuration](./packages/react-native/template/metro.config.js) that allows Metro to resolve application source files with TV-specific code, indicated by specific file extensions (e.g. `*.ios.tv.tsx`, `*.android.tv.tsx`, `*.tv.tsx`).
--- a/README.md
+++ b/README.md
@@ -94,9 +94,9 @@ See also the [Building Expo apps for TV](https://docs.expo.dev/guides/building-f
 
 ## _(New)_ How to support TV specific file extensions
 
-The template contains an [example Metro configuration](./packages/react-native/template/metro.config.js) that allows Metro to resolve application source files with TV-specific code, indicated by specific file extensions (e.g. `*.ios.tv.tsx`, `*.android.tv.tsx`, `*.tv.tsx`).
+The template contains an [example Metro configuration](./packages/react-native/template/metro.config.js) that allows Metro to resolve application source files with TV-specific code, indicated by specific file extensions (e.g. `*.ios.tv.tsx`, `*.android.tv.tsx`, `*.tv.tsx`). The config will work the same way with the other standard source file extensions (`.js`, etc.), as documented in [Metro docs](https://metrobundler.dev/docs/configuration/#sourceexts)
 
-When this is enabled, Metro will resolve files in the following order of preference:
+When this is enabled, Metro will resolve files in the following order of preference (and similarly for the other supported file extensions):
 
 - `file.ios.tv.tsx` or `file.android.tv.tsx`
 - `file.tv.tsx`
--- a/README.md
+++ b/README.md
@@ -74,23 +74,9 @@ See [this document](https://docs.expo.dev/bare/using-expo-cli/) for more details
 
 ## _(New)_ Using the Expo SDK with TV apps
 
-Starting with the Expo SDK 50 preview, and react-native-tvos 0.73.x, it will be possible to create Expo apps, and build them for TV via a new config plugin.
+See the [Building Expo apps for TV](https://docs.expo.dev/guides/building-for-tv/) guide from Expo for details, including supported Expo modules and limitations.
 
-This functionality will be new in Expo SDK 50, and will be considered an experimental feature for now.
-
-The fastest way to generate a new project is described in the [TV Example](https://github.com/expo/examples/tree/master/with-tv) in the Expo examples repo.
-
-Besides most of the core Expo modules, these also work on TV:
-
-- expo-av
-- expo-image
-- expo-localization
-- expo-updates
-
-TV does NOT support dev client (dev menu, dev launcher) at this time.
-TV does NOT support Expo Router at this time.
-
-See also the [Building Expo apps for TV](https://docs.expo.dev/guides/building-for-tv/) guide from Expo.
+Expo SDK 50 or greater, and react-native-tvos 0.73.x or later, are required.
 
 ## _(New)_ How to support TV specific file extensions
 
--- a/packages/react-native/template/metro.config.js
+++ b/packages/react-native/template/metro.config.js
@@ -1,4 +1,4 @@
-const {getDefaultConfig, mergeConfig} = require('@react-native/metro-config');
+const {getDefaultConfig} = require('@react-native/metro-config');
 
 /**
  * Metro configuration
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -90,6 +90,21 @@ Besides most of the core Expo modules, these also work on TV:
...
+## _(New)_ How to support TV specific file extensions
+
+The template contains an [example Metro configuration](./packages/react-native/template/metro.config.js) that allows Metro to resolve application source files with TV-specific code, indicated by specific file extensions (e.g. `*.ios.tv.tsx`, `*.android.tv.tsx`, `*.tv.tsx`).
+
+When this is enabled, Metro will resolve files in the following order of preference:
+
+- `file.ios.tv.tsx` or `file.android.tv.tsx`
+- `file.tv.tsx`
+- `file.ios.tsx` or `file.android.tsx`
+- `file.tsx`
```
okwasniewski, line 104: nit: I think it's also worth mentioning that not only `.tsx` files will be resolved and files can have the following extensions: `['js', 'jsx', 'json', 'ts', 'tsx']`. (https://metrobundler.dev/docs/configuration/#sourceexts)

packages/react-native/template/metro.config.js
```diff
@@ -1,11 +1,30 @@
...
+// This code is not enabled by default, since it will impact bundling performance,
+// but is available for developers who need this capability.
+//
+/*
+const originalSourceExts = config.resolver.sourceExts;
+const tvSourceExts = [
+  ...originalSourceExts.map((e) => `tv.${e}`),
+  ...originalSourceExts,
+];
+config.resolver.sourceExts = tvSourceExts;
```
okwasniewski, line 27: How would this work when bundling for mobile? If I understand this correctly users should enable this when bundling for tv?

packages/react-native/template/metro.config.js
```diff
@@ -1,11 +1,30 @@
...
 /**
  * Metro configuration
  * https://facebook.github.io/metro/docs/configuration
  *
  * @type {import('metro-config').MetroConfig}
  */
-const config = {};
+const config = getDefaultConfig(__dirname);
 
-module.exports = mergeConfig(getDefaultConfig(__dirname), config);
```
thymikee, line 11: Is this the way Metro team intends to merge configs? cc @huntie

packages/react-native/template/metro.config.js
douglowder: Correct, this should only be enabled when building for TV. In [the Expo example](https://github.com/expo/examples/blob/master/with-tv/metro.config.js) I enable it with an environment variable. I'll come up with a non-Expo-dependent way to show that here.

packages/react-native/template/metro.config.js
douglowder: In discussions with Metro experts at Expo, the general feeling is that using `mergeConfig` can be tricky, so it's safer to get the default config and just make exactly the changes you need.

thymikee: Yeah it gets the job done. At the same time I'd like the core Metro team to be aware that the community needs to "hack" the official approach

README.md
```diff
@@ -90,6 +90,21 @@ Besides most of the core Expo modules, these also work on TV:
 TV does NOT support dev client (dev menu, dev launcher) at this time.
 TV does NOT support Expo Router at this time.
 
+See also the [Building Expo apps for TV](https://docs.expo.dev/guides/building-for-tv/) guide from Expo.
+
+## _(New)_ How to support TV specific file extensions
+
+The template contains an [example Metro configuration](./packages/react-native/template/metro.config.js) that allows Metro to resolve application source files with TV-specific code, indicated by specific file extensions (e.g. `*.ios.tv.tsx`, `*.android.tv.tsx`, `*.tv.tsx`).
```
douglowder, line 97:
```suggestion
The template contains an [example Metro configuration](./packages/react-native/template/metro.config.js) that allows Metro to resolve application source files with TV-specific code, indicated by specific file extensions (e.g. `*.ios.tv.tsx`, `*.android.tv.tsx`, `*.tv.tsx`). The config will work the same way with the other standard source file extensions (`.js`, etc.), as documented in [Metro docs](https://metrobundler.dev/docs/configuration/#sourceexts)
```

README.md
```diff
@@ -90,6 +90,21 @@ Besides most of the core Expo modules, these also work on TV:
...
+
+When this is enabled, Metro will resolve files in the following order of preference:
```
douglowder, line 99:
```suggestion
When this is enabled, Metro will resolve files in the following order of preference (and similarly for the other supported file extensions):
```

packages/react-native/template/metro.config.js
douglowder, line 27:
```suggestion
if (process.env.BUILDING_FOR_TV) {
  const originalSourceExts = config.resolver.sourceExts;
  const tvSourceExts = [
    ...originalSourceExts.map((e) => `tv.${e}`),
    ...originalSourceExts,
  ];
  config.resolver.sourceExts = tvSourceExts;
}
```

README.md
douglowder: Thanks @okwasniewski I've made some changes to clarify this.

packages/react-native/template/metro.config.js
douglowder: Done.

packages/react-native/template/metro.config.js
```diff
@@ -1,11 +1,32 @@
...
+// (e.g., *.ios.tv.tsx, *.android.tv.tsx, *.tv.tsx)
+//
+// Metro will still resolve source files with standard extensions
+// as usual if TV-specific files are not found for a module.
+//
+// This code is not enabled by default, since it will impact bundling performance,
+// but is available for developers who need this capability.
+//
+/*
+if (process.env.BUILDING_FOR_TV) {
```
okwasniewski, line 22: One more question: So users should set this env variable by them selves or is it already set?

packages/react-native/template/metro.config.js
huntie: Using `mergeConfig` is totally optional — it's a convenience. However we provide the existing pattern because it lets users add their custom config inside the `config` variable as one object, avoiding procedural assignments. I would suggest this pattern be preserved for `react-native-tvos`, to avoid fragmentation.

In this case, it's close to the example we have in the [Configuring Metro](https://reactnative.dev/docs/metro#configuring-metro) docs guide. This would replace lines 4-6 from the below with your commented override.

image

Note also that `sourceExts` are later dynamically extended by `@react-native/community-cli-plugin`. This might be another location to augment the user's input Metro config.

douglowder: Thanks @huntie! I had thought that by extending `@react-native/metro-config`, we pick up all the changes done by the community plugin. Is this not correct?

packages/react-native/template/metro.config.js
matthargett: Does this mean you couldn’t have parallel builds running in the same shell environment? Is there a way to pass this is in CLI or react-native.config.js in the specific app subdir, rather than a mutable global environment variable?

yusufyildirim (approved):
Looks like everything has been discussed already. I'm just giving the green light 😄

packages/react-native/template/metro.config.js
yusufyildirim: @matthargett Yeah, sadly. We need extra Metro capabilities to make this smarter.

okwasniewski (approved):

packages/react-native/template/metro.config.js
douglowder: @matthargett you could modify the scripts in the Xcode project build phases for starting packager and for bundling. Add the new env variable there for the TV target but not for the iOS target.

packages/react-native/template/metro.config.js
```diff
@@ -1,11 +1,32 @@
...
+const {getDefaultConfig} = require('@react-native/metro-config');
 
 /**
  * Metro configuration
  * https://facebook.github.io/metro/docs/configuration
  *
  * @type {import('metro-config').MetroConfig}
  */
-const config = {};
+const config = getDefaultConfig(__dirname);
```
huntie, line 9: Clarifying my suggestion from the other comment ⬇️. Ideally, we want end users to understand they can add to the `config` object to supply their own custom config.

```suggestion
const defaultConfig = getDefaultConfig(__dirname);

const config = {
  resolver: {
    // When enabled, the optional code below will allow Metro to resolve
    // and bundle source files with TV-specific extensions
    // (e.g., *.ios.tv.tsx, *.android.tv.tsx, *.tv.tsx)
    //
    // Metro will still resolve source files with standard extensions
    // as usual if TV-specific files are not found for a module.
    //
    // This code is not enabled by default, since it will impact bundling performance,
    // but is available for developers who need this capability.
    // [].concat(
    //   process.env.BUILDING_FOR_TV
    //     ? defaultConfig.resolver.sourceExts.map((e) => `tv.${e}`)
    //     : [],
    //   defaultConfig.resolver.sourceExts,
    // ),
  },
};

module.exports = mergeConfig(defaultConfig, config);
```

packages/react-native/template/metro.config.js
matthargett: Won’t this make project conversion and docs work more difficult later? Can we send the needed PR to metro and/or CLI (and/or repack), like @okwasniewski gas been doing to enable visionOS? Is there an incontrovertibly pressing business need to deploy this as quickly as possible with this major caveat?

packages/react-native/template/metro.config.js
```diff
@@ -6,6 +6,49 @@ const {getDefaultConfig, mergeConfig} = require('@react-native/metro-config');
...
+
+// When enabled, the optional code below will allow Metro to resolve
+// and bundle source files with TV-specific extensions
+// (e.g., *.ios.tv.tsx, *.android.tv.tsx, *.tv.tsx)
+//
+// Metro will still resolve source files with standard extensions
+// as usual if TV-specific files are not found for a module.
+//
+// This code is not enabled by default, since it will impact bundling performance,
+// but is available for developers who need this capability.
```
douglowder, line 41:
```suggestion
// but is available for developers who need this capability.
// The example shown here assumes an environment variable
// "BUILDING_FOR_TV" that is set to true only when building for TV.
```

PR #653, opened 2024-02-06
fix: TVFocusGuideView focusable prop set to false prevent focusing for all subviews

dratwas:
## Summary:

Currently, disabling focus for view and its children is impossible. It is possible only by setting disabled for all touchable/pressable/buttons inside the view.

From the developer's perspective, I would love to have a prop that could be set for the parent view and will make all children not focusable.

In this PR I fixed the `focusable` prop in TVFocusGuideView that should block focus also for subviews of `TVFocusGuideView`

## Changelog:

```
- [General][Fixed] - focusable prop in TVFocusGuideView blocks focus also for its subviews
```

## Test Plan:

It should be tested on both Android TV and tvOS platforms and both architectures.

I added a Row in `TVFocusGuide autoFocus example` with this prop. The focus engine should omit this whole row as in the videos below

https://github.com/react-native-tvos/react-native-tvos/assets/16336501/07623266-94da-4f79-a8ad-2c9ed2a73bbc

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,7 @@ class Game2048 extends React.Component {
   |---|---|---|
   | destinations | any[]? | Array of `Component`s to register as destinations of the FocusGuideView |
   | autoFocus | boolean? | If true, `TVFocusGuide` will automatically manage focus for you. It will redirect the focus to the first focusable child on the first visit. It also remembers the last focused child and redirects the focus to it on the subsequent visits. `destinations` prop takes precedence over this prop when used together. |
+  | focusable | boolean? | When set to false, this view and all its subviews will be NOT focusable. |
   | trapFocus* (Up, Down, Left, Right) | Prevents focus escaping from the container for the given directions. |
   
   More information on the focus handling improvements above can be found in [this article](https://medium.com/xite-engineering/revolutionizing-focus-management-in-tv-applications-with-react-native-10ba69bd90).
--- a/packages/react-native/Libraries/Components/TV/TVFocusGuideView.js
+++ b/packages/react-native/Libraries/Components/TV/TVFocusGuideView.js
@@ -8,13 +8,14 @@
  * @format
  */
 
+import type {HostComponent} from '../../Renderer/shims/ReactNativeTypes';
+import type {ViewProps} from '../View/ViewPropTypes';
+
+import setAndForwardRef from '../../Utilities/setAndForwardRef';
+import {Commands} from '../View/ViewNativeComponent';
+
 const React = require('react');
 const ReactNative = require('react-native');
-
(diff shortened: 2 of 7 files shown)


yusufyildirim:
Thanks! I'm okay with exposing a prop for this but, I think this functionality should already be supported by `TVFocusGuideView`. Setting `tvFocusable=false!` should do the trick. I'd consider this as a `TVFocusGuideView` bug if that's not the case. Let's first discuss if fixing that bug would be sufficient for use-case or not.

dratwas:
Indeed, I concur that the desired functionality should be achieved by utilizing the `tvFocusable/tvSelectable` prop set to false. Regrettably, it appears that this approach is not yielding the expected results. Initially, I contemplated leveraging these props, but hesitated to implement changes that could break other behaviors, considering that the TVFocusGuideView inherently has the `tvSelectable` prop set to false as its default state. A logical choice given that this view is not intended to be selectable.

>Let's first discuss if fixing that bug would be sufficient for use-case or not.

Definitely, if these props would work as described then the new prop is not necessary.

Do you have any suggestions on what should I try in order to fix it in a way that is suitable for you as well?

yusufyildirim:
I certainly get your point and I partially agree with it but your assumption of `TVFocusGuideView` not being a focusable is not entirely true. It is a focusable and it gains focus. It just doesn't _usually_ retain it, it redirects. It does retain focus on focus recovery scenarios for a brief time as well. Even though if it wasn't the case, I'd still expect `TVFocusGuideView` to block focus to their descendants when it's `tvFocusable` property is set to `false`.

To simplify things up a little, I'd say everything accompanied by the focus engine during focus calculations should be (and is) considered as "focusable". Naming is hard, I'm not fully on board with your suggested prop name `disableSubviewsFocus` as well. I'd say, let's fix this bug and add a little bit of a documentation around this behavior of `TVFocusGuide` to communicate with the devs.

There's one thing we should certainly fix in terms of naming is `isTVSelectable`, `tvFocusable` and `focusable` distinction 😄 It's too confusing and I think we should  just use `focusable cc: @douglowder

yusufyildirim:
@dratwas Thanks! We don't use `focusable` prop normally, though. It's mostly preserved for mobile targets if I'm not mistaken. I think we should change that and start using `focusable` for TV platforms as well. It requires an alignment, let's discuss that first @douglowder.

dratwas:
Hey @yusufyildirim any news on this?

Perhaps it would be advantageous to adopt the currently employed properties (tvFocusable, isTVSelectable) in this pull request for smoother merging. Later on, we can refine it to utilize 'focusable' for enhanced compatibility.

What do you think? Asking because it is important for us that this error is fixed.

douglowder (approved):

PR #654, opened 2024-02-06
Fix virtualized lists

douglowder:
- Revert changes in `packages/react-native/Libraries/Lists/VirtualizedList.js` (it now refers to the `packages/virtualized-lists` implementation as RN core does)
- Package name for virtualized lists is now `@react-native-tvos/virtualized-lists`
- Change dependency in `react-native` to point to our new package


yusufyildirim (approved):
We also need a solution for metro or babel to resolve 3rd party dependency calls to our VirtualizedList package but it doesn't need to be done in this PR 🚀

PR #657, opened 2024-02-08
[Android12] Apply concurrent list patch

DrRefactor:
## Summary:

Crashes on Android 12

https://github.com/facebook/react-native/issues/37267
https://github.com/facebook/react-native/issues/33686

## Test Plan:

Test everything.
We use this patch in production for quite some time and works good, no more related crashes.

### DIFF
--- a/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedModule.java
+++ b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedModule.java
@@ -39,6 +39,9 @@ import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.LinkedBlockingQueue;
+import android.util.Log;
+import android.os.Build;
 
 /**
  * Module that exposes interface for creating and managing animated nodes on the "native" side.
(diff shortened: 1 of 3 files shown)


packages/react-native/ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedModule.java
```diff
@@ -149,7 +152,18 @@ public long getBatchNumber() {
   }
 
   private class ConcurrentOperationQueue {
-    private final Queue<UIThreadOperation> mQueue = new ConcurrentLinkedQueue<>();
+    private final Queue<UIThreadOperation> mQueue;
+
+    public ConcurrentOperationQueue(){
+      String versionAndroid = Build.VERSION.RELEASE;
+      Log.i("VERSION12", "ConcurrentOperationQueue" + versionAndroid);
```
DrRefactor, line 159: I'll make this message more meaningful and log whether custom queue implementation is used or not

DrRefactor:
@douglowder Please decide, whether this should be merged in react-native-tvos repo or we should rather wait for core repo to address the issue? I'm submitting PR here as original repo's issue is stale for quite some time...

yusufyildirim:
Thanks for the PR! I'd say this is pretty much a react-native core issue, though. It should be addressed there.

douglowder:
I agree with @yusufyildirim that this is a core issue, and also, reading through the comments in the core repo, this change could potentially have overall performance impact.

If we were to considering merging this, I think it would have to be behind a new Android feature flag that defaults to turning the feature off, so that only users who needed this would enable it.

wesvh:
@DrRefactor I tested this patch by adding it to my patch package, but I notice a blocky UI in Android 12, did it happen to you on the site where you tried it?

DrRefactor:
@wesvh not at all. BTW, did you enable building from sources after applying this patch? If not, then it has no effect.
We build RN-TVOS with this patch and push to local maven, because building from sources did not work for us in 0.72.6.

DrRefactor:
@douglowder @yusufyildirim Thank you guys for the response!

@douglowder 
> If we were to considering merging this, I think it would have to be behind a new Android feature flag that defaults to turning the feature off, so that only users who needed this would enable it.

Should I prepare PR with such feature flag or would you rather not "pollute" this repo with RN-core related fixes?

douglowder:
@DrRefactor we would prefer to hold off merging this for now, as this is replacing a nonblocking queue with a blocking one... and there is some evidence that the fix needs to be elsewhere: https://github.com/facebook/react-native/issues/37267#issuecomment-1711278146

DrRefactor:
Sure, I agree. Thanks!

PR #658, opened 2024-02-08
fix: [Apple TV] nextFocus props is not correct while set for the focused component

dratwas:
## Summary:

In this PR I added fix for an issue with `nextFocusXXX` props that are set for the already focused component.

Native FocusGuides are added to the view while the item is focused and removed while blurred. If the `nextFocusXXX` is added to an already focused item the value is not updated (or the focus guide is not added if the previous value was not set)

In my solution, I split the `enableDirectionalFocusGuides` into a few smaller methods and used them in `nextFocusXXX` prop setter.

I check if the view in which we set the property is focused and if yes I refresh the suitable focusGuide environment.

## Changelog:

```
[iOS][Fixed] - Update focus guide environment while setting nextFocus prop for already focused component
```

## Test Plan:

Change the `nextFocusXXX` prop (nextFocusRight on the video) when the item is focused and check if the destination view is correct.

**Video before**

https://github.com/react-native-tvos/react-native-tvos/assets/16336501/322ac15f-8d56-4035-a887-87f95fe2e23b

**Video after**

https://github.com/react-native-tvos/react-native-tvos/assets/16336501/1b82d57d-aa2b-4c22-8a2c-dc775910c0a3

<details>
<summary><b>Code snippet that can be pasted in examples</b></summary>

```
/**
 * Copyright (c) Facebook, Inc. and its affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 *
 * @format
 * @flow
 */

'use strict';

const React = require('react');
const ReactNative = require('react-native');

import {useRNTesterTheme} from '../../components/RNTesterTheme';

const {View, StyleSheet, Pressable, Text, TVFocusGuideView} = ReactNative;

const screenHeight = ReactNative.Dimensions.get('window').height;
const scale = screenHeight / 1080;
const width = 200 * scale;
const height = 120 * scale;

exports.framework = 'React';
exports.title = 'TV NextFocus example';
exports.description = 'NextFocus API';
exports.displayName = 'NextFocus example';
exports.examples = [
  {
    title: 'TVFocusGuide',
    render(): React.Node {
      return <TVFocusGuideExample />;
    },
  },
];

const Button = React.forwardRef((props: $FlowFixMeProps, _) => {
  const onRefAssign = ref => {
    props.onRefAssign?.(props.index, ref);
  };
  const theme = useRNTesterTheme();
  return (
    <Pressable
      onPress={props.onPress}
      nextFocusRight={props.nextFocusRight}
      style={({pressed, focused}) =>
        focused ? styles.buttonStyleFocused : styles.buttonStyle
      }
      ref={onRefAssign}>
      <Text style={[{color: theme.LinkColor}, styles.buttonText]}>
        {props.label}
      </Text>
    </Pressable>
  );
});

const TVFocusGuideExample = () => {
  const [nextFocusIndex, setNextFocusIndex] = React.useState(0);

  const refs = React.useRef({});
  const onRefAssign = (index, ref) => {
    refs.current[index] = ref;
  };

  return (
    <View style={styles.rowContainer}>
      <TVFocusGuideView autoFocus>
        <Button
          label={`Focus to ${nextFocusIndex}`}
          nextFocusRight={refs.current[nextFocusIndex]}
          onPress={() => {
            setNextFocusIndex((nextFocusIndex + 1) % 4);
          }}
        />
      </TVFocusGuideView>
      <View>
        <Button label={'Item nr 0'} onRefAssign={onRefAssign} index={0} />
        <Button label={'Item nr 1'} onRefAssign={onRefAssign} index={1} />
        <Button label={'Item nr 2'} onRefAssign={onRefAssign} index={2} />
        <Button label={'Item nr 3'} onRefAssign={onRefAssign} index={3} />
      </View>
    </View>
  );
};

const marginSize = 20 * scale;
const styles = StyleSheet.create({
  rowContainer: {
    flexDirection: 'row',
    padding: 100 * scale,
  },
  buttonText: {
    fontSize: 30 * scale,
  },
  buttonStyle: {
    marginLeft: marginSize,
    marginRight: marginSize,
    marginTop: marginSize,
    marginBottom: marginSize,
  },
  buttonStyleFocused: {
    opacity: 0.5,
    width,
    height,
    marginLeft: marginSize,
    marginRight: marginSize,
    marginTop: marginSize,
    marginBottom: marginSize,
  },
});

```

</details>

### DIFF
--- a/packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
+++ b/packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
@@ -433,6 +433,17 @@ using namespace facebook::react;
 //
 - (void)enableDirectionalFocusGuides
 {
+  [self enableNextFocusUp];
+
+  [self enableNextFocusDown];
+
+  [self enableNextFocusLeft];
+
+  [self enableNextFocusRight];
+}
+
+- (void)enableNextFocusUp
+{
   if (self->_nextFocusUp != nil) {
     if (self.focusGuideUp == nil) {
       self.focusGuideUp = [UIFocusGuide new];
@@ -446,7 +457,10 @@ using namespace facebook::react;
 
     self.focusGuideUp.preferredFocusEnvironments = @[self->_nextFocusUp];
   }
+}
 
+- (void)enableNextFocusDown
+{
   if (self->_nextFocusDown != nil) {
     if (self.focusGuideDown == nil) {
       self.focusGuideDown = [UIFocusGuide new];
@@ -460,7 +474,10 @@ using namespace facebook::react;
 
     self.focusGuideDown.preferredFocusEnvironments = @[self->_nextFocusDown];
   }
+}
 
+- (void)enableNextFocusLeft
+{
   if (self->_nextFocusLeft != nil) {
     if (self.focusGuideLeft == nil) {
       self.focusGuideLeft = [UIFocusGuide new];
@@ -474,7 +491,10 @@ using namespace facebook::react;
 
     self.focusGuideLeft.preferredFocusEnvironments = @[self->_nextFocusLeft];
   }
+}
 
+- (void)enableNextFocusRight
+{
   if (self->_nextFocusRight != nil) {
     if (self.focusGuideRight == nil) {
       self.focusGuideRight = [UIFocusGuide new];
@@ -942,6 +962,9 @@ using namespace facebook::react;
     if (newViewProps.nextFocusUp.has_value()) {
       UIView *rootView = [self containingRootView];
       _nextFocusUp = [rootView viewWithTag:newViewProps.nextFocusUp.value()];
+      if (self.isFocused) {
+        [self enableNextFocusUp];
+      }
     } else {
       _nextFocusUp = nil;
     }
@@ -951,6 +974,9 @@ using namespace facebook::react;
     if (newViewProps.nextFocusDown.has_value()) {
       UIView *rootView = [self containingRootView];
       _nextFocusDown = [rootView viewWithTag:newViewProps.nextFocusDown.value()];
+      if (self.isFocused) {
+        [self enableNextFocusDown];
+      }
     } else {
       _nextFocusDown = nil;
     }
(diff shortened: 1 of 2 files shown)


packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
```diff
@@ -432,6 +432,17 @@ - (BOOL)canBecomeFocused
...
 //
 - (void)enableDirectionalFocusGuides
+{
+  [self enableNextFocusUp];
+
+  [self enableNextFocusDown];
+
+  [self enableNextFocusLeft];
+
+  [self enableNextFocusRight];
```
yusufyildirim, line 442: I'd personally delegate the responsibility of checking if the view is focused or not to this function. Effectively making it defensive. It'd protect us from accidentally calling it without the check in some other places

packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
```diff
@@ -969,6 +998,9 @@ - (void)updateProps:(const Props::Shared &)props oldProps:(const Props::Shared &
     if (newViewProps.nextFocusRight.has_value()) {
       UIView *rootView = [self containingRootView];
       _nextFocusRight = [rootView viewWithTag:newViewProps.nextFocusRight.value()];
+      if (self.isFocused) {
+        [self enableNextFocusRight];
+      }
```
yusufyildirim, line 1003: I'd just call the `enableDirectionalFocusGuides` wherever I need it combined with the way I described above. So I wouldn't create separate `enableNext*` functions. You'd make this PR quite a bit shorter that way.

yusufyildirim (commented):
Thanks for the PR! I left some comments but I'm also totally ok with the current implementation. Leaving the decision to @douglowder. One question, though. There's no Android change, is that how it already works?

packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
douglowder: Agreed

douglowder (changes requested):
Looks good! Before merging, I would like to see the changes requested by @yusufyildirim in [this comment](https://github.com/react-native-tvos/react-native-tvos/pull/658/files#r1486363985)

dratwas:
Hey @yusufyildirim and @douglowder thanks for the review! I applied the requested changes and added an example. I also added a video for Android. As mentioned in description - Android works fine w/o any changes ;)

PR #648, opened 2024-02-01
feat: [Apple TV]  add remote arrow buttons support to TVTextScrollView

dratwas:
## Summary:

This pull request introduces enhanced functionality to the TVTextScrollView on the tvOS platform by incorporating support for remote arrow controls. Currently, the TVTextScrollView exclusively relies on swipe gestures, a divergence from the functionality on the Android platform. With this update, users will now be able to seamlessly scroll through text within the ScrollView using remote buttons on both platforms.

Additionally, this pull request addresses a specific issue related to scrolling when there are other focusable components surrounding the TVTextScrollView. The video demonstration revealed an impediment to vertically scrolling the text when buttons were positioned below the ScrollView. Instead of allowing smooth scrolling, the focus shifted to the button below. This update rectifies this issue, ensuring a more intuitive and user-friendly scrolling experience in such scenarios.

https://github.com/react-native-tvos/react-native-tvos/assets/16336501/017350a0-2003-43e3-86c0-16fc78299791

## Changelog:
```
- [tvOS] [Added] - add support for arrow remote buttons in TVTextScrollView component
- [tvOS] [Fixed] - fix scrolling of TVTextScrollView when there are other focusable components surrounding it
```

## Test Plan:

There are two things to check in this PR.

### 1. Check if a remote arrow buttons scroll the TVTextScrollView content

**Before**

https://github.com/react-native-tvos/react-native-tvos/assets/16336501/5578b807-64ec-4a25-895b-25d751a5c043

**After**

https://github.com/react-native-tvos/react-native-tvos/assets/16336501/acc356f3-3e0f-42f6-801d-7454be12c62d

### 2. Check if scrolling is possible when there are other focusable components around

**Before**

https://github.com/react-native-tvos/react-native-tvos/assets/16336501/80c358e8-f24a-4e6d-a875-ce280228249f

**After**

https://github.com/react-native-tvos/react-native-tvos/assets/16336501/1ed29595-0b25-4b9a-8f1b-8a15bb4aff62

## Additional note
I couldn't find support for TVTextScrollView in the Fabric ScrollView component. We (@callstack) will add it soon ;)

PR #673, opened 2024-02-27
feat: implement <TVTextScrollView /> for New Architecture

mateusz1913:
## Summary:

When upstreaming [TVTextScrollView patch](https://github.com/react-native-tvos/react-native-tvos/pull/648) last month, we found that TVTextScrollView was not implemented at all for new architecture.

This PR adds swipe and focus handling in `RCTScrollViewComponentView` class (similar to `RCTScrollView`)

## Changelog:

- [tvOS] [Fixed] - make TVTextScrollView component working when New Architecture is enabled

## Test Plan:

1. Go to tvOS `rn-tester` project and install pods with `RCT_NEW_ARCH_ENABLED=1` env
2. Run the RNTester app and go to `TVTextScrollViewExample`
3. Check that scrollviews can be scrolled using tv remote with swipes or dpad buttons


douglowder:
Thank you VERY much for this! I'll test this out and then get it merged.

Issue #687, opened 2024-04-02
tvOS TVEventHandler eventKeyAction undefined


cgoldsby:
### Description

On tvOS, the TVEventHandler only sends the key name and no associated action like "key up." The event should send an `eventKeyAction` to match the Android event handler. For tvOS it will always be 1, for key up.

This suggestion is to have the payloads for key events be as similar as possible between Android and tvOS, within reason, of course. Ideally, it would be nice to have key up and down events in tvOS, similar to Android. However, this would require abandoning some of the gesture recognizers. 

I've read a few issues regarding "key down" (#233, #251). Sending an `eventKeyAction` for tvOS on "key up" may cause a regression for any codebase that checks that `eventKeyAction` is undefined.

I would love to hear everyone's feedback on this topic.

### React Native Version

0.73.6-1

### Output of `npx react-native info`

-

### Steps to reproduce

-

### Snack, screenshot, or link to a repository

-

PR #689, opened 2024-04-03
fix: Add eventKeyAction to Apple NavigationEventNotification

cgoldsby:
## Summary:
Proposed fix for #687.

On tvOS, the `NavigationEventNotification` will post an `eventKeyAction` value on the payload. The intent is to send data similar to that of Android. Android's [getAction](https://developer.android.com/reference/android/view/KeyEvent#getAction()) sends an `eventKeyAction` with a value of 0 for key down and 1 for key up.

The fix also helps identify the two events being sent for long gestures. Currently, there is no way to identify a long gesture event as an up or down action.

## Changelog:

[IOS][ADDED] - Added up and down`eventKeyAction` to tvOS navigation event notifications

## Test Plan:
After creating a TestApp, I logged events to the console and verified the `eventKeyAction`.

```javascript
  const myTVEventHandler = evt => {
    console.log(evt);
  };

  useTVEventHandler(myTVEventHandler);
  ```

### DIFF
--- a/packages/react-native/React/Base/RCTTVRemoteHandler.m
+++ b/packages/react-native/React/Base/RCTTVRemoteHandler.m
@@ -15,6 +15,7 @@
 #import "RCTLog.h"
 #import "RCTRootView.h"
 #import "RCTTVNavigationEventEmitter.h"
+#import "RCTTVNavigationEventNotification.h"
 #import "RCTUIManager.h"
 #import "RCTUtils.h"
 #import "RCTView.h"
@@ -29,32 +30,6 @@ NSString *const RCTTVDisablePanGestureNotification = @"RCTTVDisablePanGestureNot
 NSString *const RCTTVEnableGestureHandlersCancelTouchesNotification = @"RCTTVEnableGestureHandlersCancelTouchesNotification";
 NSString *const RCTTVDisableGestureHandlersCancelTouchesNotification = @"RCTTVDisableGestureHandlersCancelTouchesNotification";
 
-NSString *const RCTTVRemoteEventMenu = @"menu";
-NSString *const RCTTVRemoteEventPlayPause = @"playPause";
-NSString *const RCTTVRemoteEventSelect = @"select";
-
-NSString *const RCTTVRemoteEventLongPlayPause = @"longPlayPause";
-NSString *const RCTTVRemoteEventLongSelect = @"longSelect";
-NSString *const RCTTVRemoteEventLongUp = @"longUp";
-NSString *const RCTTVRemoteEventLongDown = @"longDown";
-NSString *const RCTTVRemoteEventLongLeft = @"longLeft";
-NSString *const RCTTVRemoteEventLongRight = @"longRight";
-
-NSString *const RCTTVRemoteEventLeft = @"left";
-NSString *const RCTTVRemoteEventRight = @"right";
-NSString *const RCTTVRemoteEventUp = @"up";
-NSString *const RCTTVRemoteEventDown = @"down";
-
-NSString *const RCTTVRemoteEventPageUp = @"pageUp";
-NSString *const RCTTVRemoteEventPageDown = @"pageDown";
-
-NSString *const RCTTVRemoteEventSwipeLeft = @"swipeLeft";
-NSString *const RCTTVRemoteEventSwipeRight = @"swipeRight";
-NSString *const RCTTVRemoteEventSwipeUp = @"swipeUp";
-NSString *const RCTTVRemoteEventSwipeDown = @"swipeDown";
-
-NSString *const RCTTVRemoteEventPan = @"pan";
-
 @interface RCTTVRemoteHandler()
 
 @property (nonatomic, copy, readonly) NSDictionary<NSString *, UIGestureRecognizer *> *tvRemoteGestureRecognizers;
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by douglowder** [comment]
I built a test app at https://github.com/douglowder/PressableExample that includes this patch, to try this out.  The key actions look correct for most of the events, however longSelect and select have no key actions. Is this going to be a problem?

File: packages/react-native/React/Base/RCTTVRemoteHandler.m [hunk 1]
```diff
@@ -475,12 +502,15 @@ - (void)panned:(UIPanGestureRecognizer *)gesture {
     CGPoint velocity = [gesture velocityInView:rootView];
 
     if (gestureState) {
-        [self sendAppleTVEvent:RCTTVRemoteEventPan withBody:@{@"state": gestureState,
-                                                              @"x": [NSNumber numberWithInt:translation.x],
-                                                              @"y": [NSNumber numberWithInt:translation.y],
-                                                              @"velocityX": [NSNumber numberWithFloat:velocity.x],
-                                                              @"velocityY": [NSNumber numberWithFloat:velocity.y],
-                                                            }];
+        NSNumber *eventKeyAction = [self eventKeyActionForGestureRecognizer:gesture];
```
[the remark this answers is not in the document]
cgoldsby: Swipe gestures also fall into this category. Since tvOS supports both press and touch gestures, I struggled to decide whether to assign an event key action to all gestures or just "Press" gestures.

Now, with your feedback, I can see how it may be better to keep the spirit of an event *key* to physical presses on the remote.

What do you think about sending the `eventKeyAction` only gestures that use `allowedPressTypes', i.e., UITapGestureRecognizer and UILongPressGestureRecognizer? And, for swipe and pan gestures, there would be no `eventKeyAction`, but the body would contain the state string.

cgoldsby:
> I built a test app at https://github.com/douglowder/PressableExample that includes this patch, to try this out. The key actions look correct for most of the events, however longSelect and select have no key actions. Is this going to be a problem?

Yea, that would be a problem. In my little app I see `eventKeyAction` for longSelect and select. I'll checkout your repo now.

```
 LOG  {"eventKeyAction": 1, "eventType": "select"}
 LOG  {"eventKeyAction": 0, "eventType": "longSelect"}
 LOG  {"eventKeyAction": 1, "eventType": "longSelect"}
 ```

**Comment by cgoldsby** [comment]
> ~I'll checkout your Example app now.~ I was able to reproduce the issue reported with your Example, looking into it..

Ahh, I see the problem. Some additional gestures in the code are posting RCTTVNavigationEventNotification. I will fix this soon (more likely tomorrow).

File: packages/react-native/React/Base/RCTTVRemoteHandler.m (hunk 1 above)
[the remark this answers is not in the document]
Comment (douglowder) [reply packages/react-native/React/Base/RCTTVRemoteHandler.m:505]: Sounds good to me!

File: packages/react-native/React/Base/RCTTVRemoteHandler.m [hunk 2]
```diff
@@ -370,117 +345,117 @@ - (void)disableGestureHandlersCancelTouches
...
 - (void)longPlayPausePressed:(UIGestureRecognizer *)r
 {
-  [self sendAppleTVEvent:RCTTVRemoteEventLongPlayPause];
+    [[NSNotificationCenter defaultCenter] postNavigationPressEventWithType:RCTTVRemoteEventLongPlayPause keyAction:r.eventKeyAction tag:nil target:nil];
 
 #if RCT_DEV
-  // If shake to show is enabled on device, use long play/pause event to show dev menu
-  [[NSNotificationCenter defaultCenter] postNotificationName:@"RCTShowDevMenuNotification" object:nil];
+    // If shake to show is enabled on device, use long play/pause event to show dev menu
+    [[NSNotificationCenter defaultCenter] postNotificationName:@"RCTShowDevMenuNotification" object:nil];
```
cgoldsby, line 367: I noticed spacing is inconsistent between files. I would be happy to update any changed lines in this PR to the desired format; if I knew what it was. ;)

packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
```diff
@@ -237,35 +238,25 @@ - (void)setFocusDestinations:(NSArray*)destinations
 
 - (void)sendFocusNotification:(__unused UIFocusUpdateContext *)context
 {
-    [self sendNotificationWithEventType:@"focus"];
+    [[NSNotificationCenter defaultCenter] postNavigationFocusEventWithTag:@(self.tag) target:@(self.tag)];
```
cgoldsby, line 241: I not sure I 100% understand event tag and target.

On Android, some ReactAndroidHWInputDeviceHelpers send both [tag and target](https://github.com/react-native-tvos/react-native-tvos/blob/92f73d1d7966dd41e3d8d9101a4e298e9462887b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/modules/core/ReactAndroidHWInputDeviceHelper.java#L199-L200), while another sends just a [tag](https://github.com/react-native-tvos/react-native-tvos/blob/92f73d1d7966dd41e3d8d9101a4e298e9462887b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/ReactAndroidHWInputDeviceHelper.java#L94).

The tag and target always have the same value.

cgoldsby (commented):
Hi @douglowder, I did a little bit of refactoring by grouping common constants into a Enum Types and using them across the code base to replace Strings.

packages/react-native/React/Base/RCTTVRemoteHandler.m
douglowder: I'll take a look and flag any issues 😄

packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
douglowder: We have a plan to clean that up.

douglowder (approved):

PR #690, opened 2024-04-04
fix: Remove spurious onClick events for Android TV

douglowder:
## Summary:

There has been special code in the `Touchable` components to ensure that Android TV does not fire duplicate `onPress` events for the same press of the DPad.

This problem has now been traced to the root cause, which was `onPress` events fired with empty events from the `onClick` handler in `Pressability`.

These have been suppressed for TV, and the Android special handling removed from the `Touchable` classes.

## Changelog:

[Android][Fixed] Remove spurious onClick events for Android TV

## Test Plan:

The `TVEventHandlerExample` in RNTester has been modernized and expanded to show remote control events as well as events from `Touchable` and `Pressable` components.

### DIFF
--- a/packages/react-native/Libraries/Components/Pressable/Pressable.js
+++ b/packages/react-native/Libraries/Components/Pressable/Pressable.js
@@ -388,19 +388,15 @@ function Pressable(props: Props, forwardedRef): React.Node {
           setFocused(false);
         }
       }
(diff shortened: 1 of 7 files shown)


douglowder:
@yusufyildirim I'm going to go ahead and merge this, it is testing well and we can revert or adjust later if needed.

PR #715, opened 2024-04-26
feat(Touchable): Implement slim version of Touchable component, dedicated for TVs, to improve performance

Zahoq:
## Summary:

Goal of the PR is to introduce new `TouchableTV` component that is going to be a slim version of `TouchableOpacity` to maximise the performance, especially on low end devices. Two main things are removed from the component:
- `animations` - opacity change upon press may be not required, so there's no need to waste the resources on such animation, especially if it's replaced by some custom one. Let's allow for user of the component to decide if the animation is required.
- `pressability` - logic was designed for the mobiles and touch interactions, it's seems to be redundant for TVs, so we can save some CPU cycles by removing `componentDidUpdate` and `pressability` related parts.

I've noticed focus movement transition improvements, depending on scenario, area of application and complexity of the components:
~ up to 0.11 s (40% of total execution time) when moving focus once between touchable elements,

## Changelog:

[GENERAL] [ADDED] - Implement slim version of Touchable component, dedicated for TVs, to improve performance

## Test Plan:

Verify that `TouchableTV` covers the same functionality as `TouchableOpacity` on TV devices, but without animations.

### DIFF
--- a/.github/workflow-scripts/addDescriptiveLabels.js
+++ b/.github/workflow-scripts/addDescriptiveLabels.js
@@ -83,6 +83,7 @@ const components = [
   'TouchableHighlight',
   'TouchableNativeFeedback',
   'TouchableOpacity',
+  'TouchableTV',
   'TouchableWithoutFeedback',
   'ViewPagerAndroid',
   'VirtualizedList',
--- a/packages/react-native-babel-preset/src/configs/lazy-imports.js
+++ b/packages/react-native-babel-preset/src/configs/lazy-imports.js
@@ -42,6 +42,7 @@ module.exports = new Set([
   'TouchableHighlight',
   'TouchableNativeFeedback',
   'TouchableOpacity',
+  'TouchableTV',
   'TouchableWithoutFeedback',
   'View',
   'VirtualizedList',
--- a/packages/react-native/Libraries/Components/Touchable/TouchableTV.d.ts
+++ b/packages/react-native/Libraries/Components/Touchable/TouchableTV.d.ts
@@ -1,0 +1,55 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ */
+
+import type * as React from 'react';
+import {Constructor} from '../../../types/private/Utilities';
+import {TimerMixin} from '../../../types/private/TimerMixin';
+import {NativeMethods} from '../../../types/public/ReactNativeTypes';
+import {TVParallaxProperties} from '../View/ViewPropTypes';
+import {TouchableMixin} from './Touchable';
+import {TouchableWithoutFeedbackProps} from './TouchableWithoutFeedback';
+import type {TVProps} from './TouchableOpacity';
+
+/**
+ * @see https://reactnative.dev/docs/touchableopacity#props
+ */
+export interface TouchableTVProps
+  extends TouchableWithoutFeedbackProps,
+    TVProps {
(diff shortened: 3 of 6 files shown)


Zahoq:
@douglowder @yusufyildirim please consider this PR as a proposal and invitation for a discussion. We've noticed that we can squeeze some performance gains by removing not used parts of `TouchableOpacity` which we were using for focus management on TV devices. We've also attempted to use `TouchableWithoutFeedback` instead of `TouchableOpacity` however due to its cloning nature:
- gains weren't as noticeable;
- it would require us to reimplement all components using Touchable component to have an additional `View`, It's not so convenient to replace TouchableOpacity with TouchableWIthoutFeedback;

cc: @dratwas, thanks for consultations!

yusufyildirim:
@Zahoq Thanks for the PR! I approach to this problem from a different perspective. I personally consider `Touchable` component family as deprecated and I wouldn't invest in them. [Pressable](https://reactnative.dev/docs/pressable) is the successor and it does address some of the things you're aiming at, animations for example.

I would also try as much as possible to improve existing components instead of introducing new ones unless we're building something completely new. The more things are aligned with the upstream core, the better.

That being said, I firmly believe that there are things we can do to improve `Pressable` component and its overall performance. If you feel like you'd like to take a look into that, feel free to do so!

Zahoq:
@yusufyildirim `Pressable` indeed doesn't rely on the `Animated.View` component, however it introduces `focused` state, that may be redundant and impacts the performance of focus management.

I'm thinking about the case where we have a component like below:
```tsx
            <>
              <Pressable onFocus={onFocus} onBlur={onBlur}>
                {focused ? <SomeIcon /> : null}
              </Pressable>
              {focused ? <SomeDescription /> : null}
            </>
```

For the such use-case of `Focusable` element, there's no need to rerender `Pressable` and store the `focused` state inside it, we need the `focused` state in the parent component. Since `Pressable` is memoised, its rerender could be entirely skipped.
Same goes for the `pressed` state, it may be not needed.

That's why I was thinking about exposing the thinest version of `Focusable` component for TVs, so we can build up features on top of it if necessary. Maybe `Pressable` could be built on top of such `SlimPressable` instead on plain `View` to add its features?

Also, as `Pressability` may be not needed, it would be great to have a version on component that doesn't rely on it.
wdyt?

yusufyildirim:
@Zahoq I share the same feelings with you about having a focus state inside Pressable but it's super easy to fix. The state is only there to support [state based styling](https://github.com/react-native-tvos/react-native-tvos/blob/tvos-v0.74.0/packages/react-native/Libraries/Components/Pressable/Pressable.js#L433) (well, maybe a tiny bit more complex than that). It's a cool concept but I doubt anybody use it because it's not well-known. We can easily eliminate the state update if the style object is not a function (thus, the update is not needed) and problem should be solved. I always wanted to apply the patch I described but its impact is not that big so I never bothered.

`Pressable` is fast enough I'd say. I'd try to improve its speed but I wouldn't introduce a new component. You can easily do that in the user land if you find it absolutely necessary. @douglowder can chime in

Zahoq:
`children` can be also a function, but ye, I agree, let me prepare a PR that skips state updates if both `style` and `children` are not functions. Additionally as a separate optional commit I'll try to separate Pressable logic into 3 components:
- shared baseline
- Mobile
- TV
that way we won't have to apply Mobile listeners on TV devices

Zahoq:
Declining it for the sake of:
https://github.com/react-native-tvos/react-native-tvos/pull/724
I've not split the component into Mobile and TV version, as the `PressIn` and `PressOut` may be used on both groups, so the `Pressability` cannot be removed.

PR #724, opened 2024-05-08
 feat(Pressable): remove redundant 'pressed' and 'focused' state updates, if states are not used at all

Zahoq:
## Summary:

Goal of this PR is to optimise `Pressable` component. `Pressable` `style` and `children` properties can, but doesn't have to be functions. Usually we passing objects or arrays. `focused` and `pressed` state are used only when `style` or `children` are `functions`, so let's update those states only in such case, otherwise let's skip state updates to improve the performance.

 That way we won't have to rerender the component when it gains focus or is being pressed (assuming that `style` and `children` are not going to be functions)

## Changelog:

[GENERAL] [CHANGED] - Improve performance of `Pressable` component.

## Test Plan:

Verify that `Pressable` updates its `focused` and `pressed`  when `style` or `children` are functions or objects/arrays.

### DIFF
--- a/packages/react-native/Libraries/Components/Pressable/Pressable.js
+++ b/packages/react-native/Libraries/Components/Pressable/Pressable.js
@@ -271,6 +271,10 @@ function Pressable(props: Props, forwardedRef): React.Node {
 
   const android_rippleConfig = useAndroidRippleForView(android_ripple, viewRef);
 
+  const isStyleAFunction = typeof style === 'function';
+  const isChildrenAFunction = typeof children === 'function';
+  const shouldUpdateStates = isStyleAFunction || isChildrenAFunction;
+
   const [pressed, setPressed] = usePressState(testOnly_pressed === true);
 
   const [focused, setFocused] = useState(false);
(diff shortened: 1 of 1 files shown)


Zahoq:
@douglowder @yusufyildirim I've created this PR as a follow up of the discussion that took place in https://github.com/react-native-tvos/react-native-tvos/pull/715
please share your thoughts

yusufyildirim:
Seems good, gotta pull the branch and test it before approving. What do you feel about this one @douglowder?

douglowder:
I fully support this, seems like an excellent performance improvement!

Zahoq:
Simliar PR I've created was merged to `react-native` repository:
https://github.com/facebook/react-native/pull/44615

douglowder:
@Zahoq I've cherry-picked your core commit (this will avoid conflicts when merging from core for picking up 0.74 fixes).

Please check https://github.com/react-native-tvos/react-native-tvos/commit/3f71fa8c61087d34106ac5a1ad435eab8ba20e72 and make sure that works correctly for you.

Thanks again for this performance improvement!

PR #725, opened 2024-05-09
fix(Android TV): Modal key and focus events with bridgeless

douglowder:
## Summary:

After 0.74.1-0, the Modal component would cause a crash when an Android TV app was running in bridgeless mode and an arrow key was pressed.

This fixes the issue by ensuring the correct context is used when emitting events.

## Changelog:

[Android][Fixed] hardware key events in Modal and bridgeless

## Test Plan:

- RNTester Modal example could reproduce the issue. After the fix, the issue is no longer reproducible.
- Test with TVInputDemo with and without Fabric and bridgeless.


douglowder:
Sadly, this works in RNTester, but does not work for modals in some other apps. So I'm closing this PR and will pick this work up later.

The root cause is discussed in https://github.com/reactwg/react-native-new-architecture/discussions/186 and is being worked on by the RN core team.

PR #766, opened 2024-08-13
RNTester tvOS fixes

dcordero:
## Summary:

Minor fixes have been made to the RNTester app for tvOS. These changes do not affect Android TV.

## Changelog:

[FIXED] - Missing LaunchScreen on RNTester for tvOS
[FIXED] - Mobile back button should not be shown in tvOS Navigation Bar.
[FIXED] - Wrong Hermes version reference from Podfile.lock file

### DIFF
--- a/packages/rn-tester/RNTester/Info.plist
+++ b/packages/rn-tester/RNTester/Info.plist
@@ -61,5 +61,7 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+	<key>UILaunchStoryboardName</key>
+	<string>LaunchScreen</string>
 </dict>
 </plist>
--- a/packages/rn-tester/js/components/RNTTitleBar.js
+++ b/packages/rn-tester/js/components/RNTTitleBar.js
@@ -43,7 +43,7 @@ const HeaderIOS = ({
             <RNTesterDocumentationURL documentationURL={documentationURL} />
           )}
         </View>
-        {onBack != null && (
+        {!Platform.isTV && onBack != null && (
           <View>
             <Button
               title="Back"
--- a/packages/rn-tester/Podfile.lock
+++ b/packages/rn-tester/Podfile.lock
@@ -1,13 +1,13 @@
 PODS:
   - boost (1.83.0)
   - DoubleConversion (1.1.6)
-  - FBLazyVector (0.74.3-0)
+  - FBLazyVector (0.74.5-0)
   - fmt (9.1.0)
   - glog (0.3.5)
-  - hermes-engine (0.74.3-0):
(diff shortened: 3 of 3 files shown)


packages/rn-tester/RNTester/Info.plist
```diff
@@ -61,5 +61,7 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+	<key>UILaunchStoryboardName</key>
+	<string>LaunchScreen</string>
```
dcordero, line 65: LaunchScreen

packages/rn-tester/js/components/RNTTitleBar.js
```diff
@@ -43,7 +43,7 @@ const HeaderIOS = ({
             <RNTesterDocumentationURL documentationURL={documentationURL} />
           )}
         </View>
-        {onBack != null && (
+        {!Platform.isTV && onBack != null && (
```
douglowder, line 46: This change would also affect Android TV, as `Platform.isTV` is also true there.

douglowder (changes requested):
Why does the Back button need to be removed for Apple TV?

We can't merge this change at the moment because we are about to release 0.75 version. Once that is done, please rebase this PR against the tvos-v0.75.0 branch.

The `Podfile.lock` file is updated as part of the release process, so you should remove that from the PR. We will take a look to ensure that it gets updated properly during the release generation.

packages/rn-tester/js/components/RNTTitleBar.js
dcordero: I think that this shouldn't be the case, as this condition is part of the definition of `HeaderIOS`.

The Android case is handled by a different component, `HeaderAndroid`, which is defined directly below this one."

dcordero:
> Why does the Back button need to be removed for Apple TV?

On-screen Back buttons are specific to iOS. Big screen apps typically handle back navigation using the back key on the remote control.

Similarly, the RNTester App does not have already an on-screen back button for Android mobile, as this platform relies on the standard Android back button for navigation.

dcordero:
> The `Podfile.lock` file is updated as part of the release process, so you should remove that from the PR

Thanks for the heads up. I have reverted the changes on Podfile.lock

dcordero:
> We can't merge this change at the moment because we are about to release 0.75 version. Once that is done, please rebase this PR against the tvos-v0.75.0 branch.

Thanks for reviewing @douglowder.  I will rebase and adjust this code to the latest changes. I saw this code has been actually affected by 0.75.0 release with [[this commit](https://github.com/react-native-tvos/react-native-tvos/commit/712ff8cdba59bd5f8bb1790f60171e0f126e415c)].

PR #791, opened 2024-09-28
feat: Add NativeWind support to Pressable

douglowder:
TV controls such as `Touchable*` and `Pressable` already support the `focus:` NativeWind style modifier. This change allows `Pressable` to support the `active:` NativeWind style modifier on TV.

See https://www.nativewind.dev/v4/core-concepts/states#hover-focus-and-active-

(This does not add `active:` support for the `Touchable*` components.)

### DIFF
--- a/packages/react-native/Libraries/Components/Pressable/Pressable.js
+++ b/packages/react-native/Libraries/Components/Pressable/Pressable.js
@@ -408,12 +408,17 @@ function Pressable(
       // $FlowFixMe[prop-missing]
       if (evt.eventType === 'select') {
         // $FlowFixMe[incompatible-exact]
+        onPressIn(evt);
         onPress && onPress(evt);
+        setTimeout(() => {
+          onPressOut(evt)
+        }, unstable_pressDelay ?? 100);
       }
       // $FlowFixMe[prop-missing]
       if (evt.eventType === 'longSelect') {
         // $FlowFixMe[incompatible-exact]
-        onLongPress && onLongPress(evt);
+        evt.eventKeyAction === 0 ? onPressIn(evt) : onPressOut(evt);
+        // onLongPress && onLongPress(evt);
       }
     },
     [


douglowder:
@marklawlor FYI

douglowder:
@yusufyildirim for a demo, see https://github.com/douglowder/NativeWindTV/tree/doug/nativewind-v4

packages/react-native/Libraries/Components/Pressable/Pressable.js
```diff
@@ -408,12 +408,17 @@ function Pressable(
       // $FlowFixMe[prop-missing]
       if (evt.eventType === 'select') {
         // $FlowFixMe[incompatible-exact]
+        onPressIn(evt);
```
XITE-yusufyildirim, line 411: Sorry for the late review but I wonder if we should solve this problem on the native layer by sending the expected events for [Pressability](https://github.com/react-native-tvos/react-native-tvos/blob/tvos-v0.76.0/packages/react-native/Libraries/Pressability/Pressability.js) to do its job. Here we're introducing our own way of doing the same thing which is a bit confusing and redundant if you ask me

XITE-yusufyildirim (commented):
Late to the party, sorry about that 😄 I have a little concern regarding this implementation. I feel like it'd be better if we focus on making Pointer events fully functional on TV env (as much as possible) which would solve this issue and probably some others we're not aware of. 

I haven't tried but `react-strict-dom` also has `:active` state polyfill and it [depends](https://github.com/facebook/react-strict-dom/blob/main/packages/react-strict-dom/src/native/modules/usePseudoStates.js#L78) on `onPointerUp/Down` events. Haven't tried if they work on TV but I assume not based on this PR. That's why I think we should solve the bigger problem.

What do you think?

packages/react-native/Libraries/Components/Pressable/Pressable.js
douglowder: I definitely see your point. NativeWind is using its own package called `react-native-css-interop`, which is looking for `onPressIn/Out` events. Ideally, we should be trying to support all these events, and maybe even be able to bring in the `Touchable` components too if we can. I'll take another look at this to see if we can move these changes to the `Pressability` module.

douglowder: One complication that arises in trying to handle this purely from the native layer, is that on Apple TV there are no key up/key down events like in Android TV, that can be directly mapped to `onPointerUp/Down`. What you have instead are gesture recognizers that fire handlers when a select or longSelect is detected. For longSelect, you do get two events, one at the start and one at the end, so that is no problem. For just a select, you only get one event, so we would have to generate the pointer up/down events synthetically anyway....

douglowder: FYI @yusufyildirim I went ahead and reverted this change in tvos-v0.76.0 branch and in the 0.76.0-0rc3 release that will be published tomorrow.

douglowder:
> Late to the party, sorry about that 😄 I have a little concern regarding this implementation. I feel like it'd be better if we focus on making Pointer events fully functional on TV env (as much as possible) which would solve this issue and probably some others we're not aware of.
> 
> I haven't tried but `react-strict-dom` also has `:active` state polyfill and it [depends](https://github.com/facebook/react-strict-dom/blob/main/packages/react-strict-dom/src/native/modules/usePseudoStates.js#L78) on `onPointerUp/Down` events. Haven't tried if they work on TV but I assume not based on this PR. That's why I think we should solve the bigger problem.
> 
> What do you think?

Hey @yusufyildirim I've come back to this issue, testing with this app: https://github.com/douglowder/NativeWindTV/tree/doug/nativewind-v4

I'm looking through how `Pressable` and `Pressability` work in RN 0.75 and later, and it looks like a more scalable way to do this is to either hook into `Pressability.js` itself, or possibly add  `useTVEventHandler` to the `usePressability.js` hook. Then any component using `Pressability` would get the right support for these events, and we would not have to make native code changes. I'll try some of this out and see if I can get it working before making a 0.76 PR.

PR #803, opened 2024-10-19
feat: support NativeWind, refactor Pressable for TV

douglowder:
Improved implementation of the NativeWind support added in #791 and then reverted.

- Move `Pressable` TV event handling to `Pressability` as another event handler in its collection of handlers
- Add code to call onPressIn/onPressOut for TV select/longSelect events
- Add new `tvPressDuration` prop to `Pressable` to control length of select events
- Fix unit tests for the new `View` prop added
- Stub out a bad unit test with open handle problems
- New code in `TVEventHandlerExample` in `RNTester` to exercise these features


douglowder:
@yusufyildirim with regard to your comment about supporting W3C pointer events with this change (https://github.com/react-native-tvos/react-native-tvos/pull/791#pullrequestreview-2342319037):

I think that we could probably implement this pretty easily after this change, as it only requires some additional code to fire the `onPointerDown/Up` events in the new `onTVEvent()` prop that I added here. If you agree, we can create a new PR or issue later to add that and the appropriate testing.

douglowder:
@yusufyildirim I added a new prop to `Pressable`, to allow the developer to control how long the component stays in the "pressed" state on a select event on TV platforms. Unit tests pass, I added TS type for the new prop, and added some code in the TVEventHandler example in RNTester to test that it works, show `onPressIn/Out` events, etc.

XITE-yusufyildirim (commented):
-

yusufyildirim (commented):
I think we should rethink our tv event handler approach and align it with regular RN event flows. We shouldn't need to implement anything special for this feature to work in an ideal world. If we were triggering the corresponding events from the native side just like upstream RN does, everything would work just fine. We don't do that though. We have our own separate event channel for TV remote events and we handle all that on the JS side and emulate everything there. This approach has been bogging me for a long time cause it's separating us from the upstream. Some of the core features doesn't work out of the box (like pointer events here) because of this approach.

Like, since we emulate the press events on the JS side, you have to rely on the tvPressDuration because you don't know when user lets go of the select button on their remote, right? That's a problem.

I gave it a shot to overhaul this system I think 6 months ago but dropped it due to the workload I had at the time. I'd like to give it an another shot soon. Until then, I think it's fine to go with any approach you find suitable.

douglowder:
On Apple TV, you actually do not know when the user releases the remote button -- the tvOS gesture recognizer only creates one native event, not a key down key up pair as on Android TV. 

That said, I agree that it would be worthwhile to explore doing this natively in future!

yusufyildirim:
Ah yeah, you told that before. We might be able to come up with something ourselves maybe. Gotta dig into the documentation for that 😄

douglowder:
Ok @yusufyildirim after further thought:

- Stage 1: Let's merge this
- Stage 2: A PR to move all the `Touchable` components to go through the same `Pressability` handler, so that everything goes through the same code path
- Stage 3: Rework things so that we are sending native events for onPressIn/Out directly to the tvFocusHandler and then to `Pressability`

What do you think?

yusufyildirim:
@douglowder works for me. My only concern would be the new tvPressDuration prop. We may delete it in the future and it'll be a breaking change. Adoption will be probably low, though.

douglowder:
That's a good point! I'll revert `tvPressDuration` before merging, and we can see if that is really needed in future.

PR #813, opened 2024-11-03
Implement native press and focus events

yusufyildirim:
# WIP

Implements #668.

### DIFF
--- a/packages/react-native/Libraries/Pressability/Pressability.js
+++ b/packages/react-native/Libraries/Pressability/Pressability.js
@@ -157,7 +157,8 @@ export type EventHandlers = $ReadOnly<{|
   onResponderTerminate: (event: PressEvent) => void,
   onResponderTerminationRequest: () => boolean,
   onStartShouldSetResponder: () => boolean,
-  onTVEvent: (event: any) => void,
+  onPressIn: (event: any) => void,
+  onPressOut: (event: any) => void,
 |}>;
 
 type TouchState =
@@ -397,6 +398,7 @@ export default class Pressability {
   |}>;
   _touchActivateTime: ?number;
   _touchState: TouchState = 'NOT_RESPONDER';
+  _longPressSent: boolean = false;
 
   constructor(config: PressabilityConfig) {
     this.configure(config);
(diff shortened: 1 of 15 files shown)


packages/react-native/React/Views/RCTTVView.m
```diff
@@ -123,18 +114,25 @@ - (void)handleSelect:(UIGestureRecognizer *)r
...
-- (void)handleLongSelect:(UIGestureRecognizer *)r
+- (void)handlePress:(UIGestureRecognizer *)r
 {
-    [self sendLongSelectNotification:r];
+  switch (r.state) {
+      case UIGestureRecognizerStateBegan:
+      if (self.onPressIn) self.onPressIn(nil);
+          break;
+      case UIGestureRecognizerStateEnded:
+      case UIGestureRecognizerStateCancelled:
```
douglowder, line 129: If the press is cancelled, should we be sending an onPressOut?

douglowder: Maybe that's not something that can happen with TV remote presses...

douglowder (approved):

packages/react-native/React/Views/RCTTVView.m
yusufyildirim: That's a really good question. We probably need to put some more thought into this. I'm not sure it is really possible on TV as you questioned it. Added the clause just to be safe. On the other hand, there's this case where you start the press action and navigate to a different element when the `select` button is held down. So, you start the action when being focused on an element, you end it when you're focused on a completely different one. I think we should treat that case as cancellation but I'm not sure.

That being said, I think we should send onPressOut in the cancelation too. Imagine it like a pointer event, it doesn't mean the button is `pressed`. That's why we probably shouldn't generate an `onPress` or `onLongPress` event in that case. Gotta check the platform standards (incl. web) and see how that case should be handled.

Let's consider that as an edge case and create a bug ticket to refine the implementation later.

packages/react-native/Libraries/Pressability/Pressability.js
```diff
@@ -567,6 +584,7 @@ export default class Pressability {
       onClick: (event: any): void => {
         // If event has `pointerType`, it was emitted from a PointerEvent and
         // we should ignore it to avoid triggering `onPress` twice.
+        debugger;
```
mateusz1913, line 587: 👋 this should probably be removed
cc: @yusufyildirim @douglowder

yusufyildirim: Haha, amazing. Thanks for catching that 😄

douglowder: Line has been removed in https://github.com/react-native-tvos/react-native-tvos/commit/4ae06004665d7f72147422b42d2bd65c1e192969

PR #814, opened 2024-11-04
fix: unset nextFocus guides when prop set to undefined

christianEconify:
Unset focus guides when changing the nextFocus prop back to undefined

### DIFF
--- a/packages/react-native/React/Views/RCTTVView.m
+++ b/packages/react-native/React/Views/RCTTVView.m
@@ -458,23 +458,43 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder : unused)
 }
 
 - (void)setNextFocusUp:(NSNumber *)nextFocusUp {
-  self->_nextFocusUp = [self getViewById: nextFocusUp];
-  [self enableDirectionalFocusGuides];
+  if (self.focusGuideUp != nil && nextFocusUp == nil) {
+    [[self rootView] removeLayoutGuide:self.focusGuideUp];
+    self.focusGuideUp = nil;
+  } else {
+    self->_nextFocusUp = [self getViewById: nextFocusUp];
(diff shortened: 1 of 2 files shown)


packages/react-native/React/Views/RCTTVView.m
```diff
@@ -458,23 +458,43 @@ - (RCTTVView *)getViewById:(NSNumber *)viewId {
 }
 
 - (void)setNextFocusUp:(NSNumber *)nextFocusUp {
-  self->_nextFocusUp = [self getViewById: nextFocusUp];
-  [self enableDirectionalFocusGuides];
+  if (self.focusGuideUp != nil && nextFocusUp == nil) {
```
douglowder, line 461: This change makes sense! Could you please have a look at the Fabric implementation as well, to see if similar adjustments are needed?

https://github.com/react-native-tvos/react-native-tvos/blob/13afea00a2495858266ec83416b86930d464033f/packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm#L1074

douglowder (commented):
See inline

packages/react-native/React/Views/RCTTVView.m
christianEconify: Done

christianEconify: Just a note, struggled to get nextFocus working on rn-tester fabric before and after the change - can you check it out also?

douglowder (approved):

packages/react-native/React/Views/RCTTVView.m
douglowder: Yes, there is a nextFocus example in RNTester that works fine with Fabric. I'll test this and verify on tvOS.

douglowder (approved):

PR #821, opened 2024-11-12
fix: not checking next focus down or right prop correctly

christianEconify:
Not too sure how I missed this tbh

Correctly check the prop for `nil` to unset the focus guide


douglowder (approved):

christianEconify:
Hey @douglowder 
Will a new version of 74 be released - now these changes have all been merged in?

douglowder:
We are pushing out a 0.76 release... I can see how much work it will be to push out 0.74. We generally don't go back more than one version...

christianEconify:
That's understandable - We already have these changes in our project via a patch-package. But the less patches the better haha.
Cheers for the speedy response, I have also sent you an email if you could check that out :) !

PR #824, opened 2024-11-13
Optimize androidHWInputDeviceHelper to use lazy initialization

cgoldsby:
The computed property `androidHWInputDeviceHelper` previously used a custom getter that created a new instance of `ReactAndroidHWInputDeviceHelper` each time it was accessed. This behavior led to a new instance being created with every key press, causing issues with maintaining state, such as tracking the last focused button.

By switching to a `by lazy` initialization approach, `androidHWInputDeviceHelper` is now instantiated only once upon its first access, preserving its state across key presses. This ensures that the "select" key functions correctly by retaining focus information across button interactions.


cgoldsby:
I was unable to run RNTester on the default branch. I tested with the latest tag, v0.76.1-0, and patched the fix.

[mypatch.patch](https://github.com/user-attachments/files/17739650/mypatch.patch)

```shell
gh repo clone react-native-tvos/react-native-tvos
cd react-native-tvos
git checkout v0.76.1-0
git apply /path/to/mypatch.patch
yarn ; yarn android
```

douglowder (approved):

PR #825, opened 2024-11-15
fix: restore TV event handler select/longSelect for tvOS

douglowder:
The introduction of fully native focus/blur/pressIn/pressOut in #813 led to an inconsistency in the behavior of the TV event handler. On Android TV, the `select` and `longSelect` events are still sent to the handler. On Apple TV, they are missing.

Root cause was interference between the select gesture recognizer in the views, and the one in `RCTTVRemoteHandler`.

Solution:

- Remove handling of select and longSelect from `RCTTVRemoteHandler`
- Create a new class, `RCTTVRemoteSelectHandler`, that has the select and longSelect gesture recognizers, and calls delegate methods in the views when these gestures start and finish
- The new class is used by both Fabric and non-Fabric views, to avoid code duplication.

Before:

Simulator Screenshot - Apple TV - 2024-11-15 at 00 06 35

After:

Simulator Screenshot - Apple TV - 2024-11-15 at 00 10 12

### DIFF
--- a/packages/react-native/React/Base/RCTTVRemoteSelectHandler.m
+++ b/packages/react-native/React/Base/RCTTVRemoteSelectHandler.m
@@ -1,0 +1,111 @@
+#import "RCTTVRemoteSelectHandler.h"
+
+@interface RCTTVRemoteSelectHandler()
+
+@property (nonatomic, strong) UILongPressGestureRecognizer * pressRecognizer;
+@property (nonatomic, strong) UILongPressGestureRecognizer * longPressRecognizer;
+
+@property (nonatomic, weak) UIView<RCTTVRemoteSelectHandlerDelegate> *view;
+
+@end
+
+@implementation RCTTVRemoteSelectHandler {
+  NSMutableDictionary<NSString *, UIGestureRecognizer *> *_tvRemoteGestureRecognizers;
+}
+
+#pragma mark -
+#pragma mark Public methods
+
+- (instancetype)initWithView:(UIView <RCTTVRemoteSelectHandlerDelegate> *)view
+{
+  if ((self = [super init])) {
+      _view = view;
+      [self attachToView];
+  }
+  return self;
+}
+
+- (void)dealloc
+{
+  [self detachFromView];
+  _view = nil;
+}
+
+#pragma mark -
+#pragma UIGestureRecognizerDelegate method
+
+// Press recognizer should allow long press recognizer to work (but not the reverse)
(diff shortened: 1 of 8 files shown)


packages/react-native/React/Base/RCTTVRemoteSelectHandler.m
```diff
@@ -0,0 +1,111 @@
...
+    [self.view removeGestureRecognizer:_longPressRecognizer];
+    self.longPressRecognizer = nil;
+  }
+}
+
+- (void)handlePress:(UIGestureRecognizer *)r
+{
+  switch (r.state) {
+    case UIGestureRecognizerStateBegan:
+      NSLog(@"selectGestureBegan");
```
douglowder, line 77:
```suggestion
```

packages/react-native/React/Base/RCTTVRemoteSelectHandler.m
```diff
@@ -0,0 +1,111 @@
...
+      [self.view selectGestureBegan];
+      break;
+    case UIGestureRecognizerStateCancelled:
+    case UIGestureRecognizerStateEnded:
+      if (r.enabled) {
+        NSLog(@"selectGestureEnded");
```
douglowder, line 83:
```suggestion
```

packages/react-native/React/Base/RCTTVRemoteSelectHandler.m
```diff
@@ -0,0 +1,111 @@
...
+      break;
+  }
+}
+
+- (void)handleLongPress:(UIGestureRecognizer *)r
+{
+  switch (r.state) {
+    case UIGestureRecognizerStateBegan:
+      self.pressRecognizer.enabled = NO;
+      NSLog(@"longSelectGestureBegan");
```
douglowder, line 97:
```suggestion
```

packages/react-native/React/Base/RCTTVRemoteSelectHandler.m
```diff
@@ -0,0 +1,111 @@
...
+      [self.view longSelectGestureBegan];
+      break;
+    case UIGestureRecognizerStateEnded:
+    case UIGestureRecognizerStateCancelled:
+      NSLog(@"longSelectGestureEnded");
```
douglowder, line 102:
```suggestion
```

MajStc:
Hi! It seems that this change has broke 'enter' events on tvOS 18.1, reverting to 76.1 it works fine, could You take a look at that again? :)

douglowder:
> Hi! It seems that this change has broke 'enter' events on tvOS 18.1, reverting to 76.1 it works fine, could You take a look at that again? :) 

@MajStc can you provide sample code to show the issue?

Issue #829, opened 2024-11-19
Key Events Propagation Issue on Android TV's Modals


cgoldsby:
## Summary
In our app, users can interact with the following remote control buttons:

- **Channel Up (CH+)**
- **Channel Down (CH-)**
- **Number Keys (0–9)**

For Android TVs with a TV tuner (e.g., Sony Bravia), the app should capture these Key events to prevent them from propagating to the system.

In our app, we override the `onKeyDown` and `onKeyUp` methods in `MainActivity` to consume these key events:
```kotlin
override fun onKeyDown(keyCode: Int, event: KeyEvent?): Boolean {
    when (event?.keyCode) {
        KeyEvent.KEYCODE_0, ..., KeyEvent.KEYCODE_9,
        KeyEvent.KEYCODE_CHANNEL_UP,
        KeyEvent.KEYCODE_CHANNEL_DOWN -> return true
    }
    return super.onKeyDown(keyCode, event)
}

override fun onKeyUp(keyCode: Int, event: KeyEvent?): Boolean {
    when (event?.keyCode) {
        KeyEvent.KEYCODE_0, ..., KeyEvent.KEYCODE_9,
        KeyEvent.KEYCODE_CHANNEL_UP,
        KeyEvent.KEYCODE_CHANNEL_DOWN -> return true
    }
    return super.onKeyUp(keyCode, event)
}
```

This solution works. **Except** for **React Native Modal** (i.e. Android's `Dialog`).

## Observed Issue
When a **React Native Modal** (`Dialog`) is presented, **KEY_DOWN** events are not passed to `MainActivity`. This causes the **CH+**, **CH-**, and **0–9** buttons to still propagate their Key events to the system.

When the Key events propagate to the system, the **unexpected** behavior occurs:

1. The app handles the Key event. ✅
2. The system forcefully closes the app. ❌
3. The system tunes the TV to a channel. ❌

### Why This Happens:

- In React Native, the Android Dialog's **KEY_UP** events are delegated to MainActivity.
- However, **KEY_DOWN** events are not delegated when a modal is active.
https://github.com/react-native-tvos/react-native-tvos/blob/c97452ba91f8f73fc806c63026da43bbd28d5354/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.kt#L261

## Solution
The solution is to delegate **KEY_DOWN** events to the `MainActivity`, i.e., current activity.
```diff
[...]

PR #847, opened 2024-12-20
[Release 0.77.0-0rc5] Fix Key Events Propagation Issue on Android TV's Modals #829

cgoldsby:
### Description
---
This pull request addresses the key events propagation issue reported in #829 in the `react-native-tvos` repository. The fix incorporates feedback from the original issue discussion and resolves the problem where `ACTION_DOWN` key events were not properly propagated to the Android app's main activity modals.

### Changes
---
- Added a condition to delegate `ACTION_DOWN` key events to `onDialogKeyDown` when a modal is open.
- Updated handling of `ACTION_UP` key events to be delegated to `onDialogKeyUp`.
- Tries to incorporate feedback from the original issue discussion.

### Testing Instructions
---
1. Clone the repository:
   ```bash
   gh repo clone cgoldsby/react-native-tvos
   ```
2. Checkout the branch `modalFix-test`:
   ```bash
   git checkout modalFix-test
   ```
3. Install dependencies:
   ```bash
   yarn
   ```
4. Build and run the project on Android:
   ```bash
   yarn android
   ```

5. Navigate to the "Modal" page in the RNTester app. While a modal is open, pressing up or down on any key (other than back or escape) should log a message in Android Studio's Logcat, confirming that the key events are now correctly propagated to the app's main activity.

---
### Breaking Change

The current implementation handles `ACTION_UP` by delegating to the app's activity `onKeyUp` function. This behavior has been modified. `ACTION_UP` is now delegated to `onDialogKeyUp` instead.

To retain the previous behavior, you can override `onDialogKeyUp` as follows:

```kotlin
override fun onDialogKeyUp(keyCode: Int, event: KeyEvent?): Boolean {
    return this.onKeyUp(keyCode, event)
}
```

This will ensure that `onKeyUp` is still called when the `ACTION_UP` event occurs.

---
### Thank you
Thanks for checking out this PR! 🙌 Any feedback you have would be awesome to make this change even more 🌟. Appreciate your time! 😊

### DIFF
--- a/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.kt
+++ b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.kt
@@ -29,6 +29,7 @@ import android.widget.FrameLayout
 import androidx.annotation.UiThread
 import com.facebook.common.logging.FLog
 import com.facebook.react.R
+import com.facebook.react.ReactActivity
 import com.facebook.react.bridge.GuardedRunnable
 import com.facebook.react.bridge.LifecycleEventListener
 import com.facebook.react.bridge.ReactContext
@@ -275,12 +276,31 @@ public class ReactModalHostView(context: ThemedReactContext) :
                 // activity expects to receive those events and react to them, ie. in the case of
                 // the dev menu
                 val innerCurrentActivity =
-                    (this@ReactModalHostView.context as ReactContext).currentActivity
+                    (this@ReactModalHostView.context as ReactContext).currentActivity as? ReactActivity
                 if (innerCurrentActivity != null) {
                   return innerCurrentActivity.onKeyUp(keyCode, event)
                 }
               }
             }
+
+            // @see https://github.com/react-native-tvos/react-native-tvos/issues/829
+            if (event.action == KeyEvent.ACTION_DOWN) {
+              // Allow BACK and ESCAPE keys to propagate to the dialog's activity.
+              // Returning false ensures these keys are not consumed here, so they
+              // can be processed during the ACTION_UP event where their functionality
+              // is handled by JavaScript.
+              if (keyCode == KeyEvent.KEYCODE_BACK || keyCode == KeyEvent.KEYCODE_ESCAPE) {
+                return false
+              }
+
+              // Redirect all other key down events to the current activity.
+              val innerCurrentActivity =
+                (this@ReactModalHostView.context as ReactContext).currentActivity as? ReactActivity
+              if (innerCurrentActivity != null) {
+                return innerCurrentActivity.onKeyDown(keyCode, event)
+              }
+            }
+
             return false
           }
         })
--- a/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.kt
+++ b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.kt
@@ -278,7 +278,7 @@ public class ReactModalHostView(context: ThemedReactContext) :
                 val innerCurrentActivity =
                     (this@ReactModalHostView.context as ReactContext).currentActivity as? ReactActivity
                 if (innerCurrentActivity != null) {
-                  return innerCurrentActivity.onKeyUp(keyCode, event)
+                  return innerCurrentActivity.onDialogKeyUp(keyCode, event)
                 }
               }
             }
@@ -297,7 +297,7 @@ public class ReactModalHostView(context: ThemedReactContext) :
               val innerCurrentActivity =
                 (this@ReactModalHostView.context as ReactContext).currentActivity as? ReactActivity
               if (innerCurrentActivity != null) {
-                return innerCurrentActivity.onKeyDown(keyCode, event)
+                return innerCurrentActivity.onDialogKeyDown(keyCode, event)
               }
             }
 
--- a/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/ReactActivity.java
+++ b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/ReactActivity.java
@@ -144,6 +144,14 @@ public abstract class ReactActivity extends AppCompatActivity
     mDelegate.onConfigurationChanged(newConfig);
   }
 
+  public boolean onDialogKeyUp(int keyCode, KeyEvent event) {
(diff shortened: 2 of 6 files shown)


douglowder:
This definitely looks good at first glance. Is there any way that you would be able to add a test for this in the ModalExample in RNTester?

cgoldsby:
> This definitely looks good at first glance. Is there any way that you would be able to add a test for this in the ModalExample in RNTester?

I would gladly look into adding tests. Due to the holidays, I can look into this at the beginning of the new year. Hope that works for everyone. See you then! Happy Holidays!

cgoldsby:
> > This definitely looks good at first glance. Is there any way that you would be able to add a test for this in the ModalExample in RNTester?
> 
> I would gladly look into adding tests. Due to the holidays, I can look into this at the beginning of the new year. Hope that works for everyone. See you then! Happy Holidays!

Hi @douglowder, I made an effort to add a test to the RNTester app but encountered a limitation with tvos-76.0-5. 

Specifically, the reactContext is not accessible within an activity when using the new architecture. https://github.com/facebook/react-native/issues/47647

Fortunately, this limitation has been addressed in v0.77.0. As a result, I rebased my branch onto release-0.77.0-0rc5 to leverage the fix.

In the RNTester app, I added an example that displays the key up and down events propagated to the main activity.

[untitled.webm](https://github.com/user-attachments/assets/87d1f97d-df50-4ac1-9d42-f51868c27410)

douglowder (approved):

packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.kt
```diff
@@ -275,12 +276,31 @@ public class ReactModalHostView(context: ThemedReactContext) :
...
+
+            // @see https://github.com/react-native-tvos/react-native-tvos/issues/829
+            if (event.action == KeyEvent.ACTION_DOWN) {
+              // Allow BACK and ESCAPE keys to propagate to the dialog's activity.
+              // Returning false ensures these keys are not consumed here, so they
+              // can be processed during the ACTION_UP event where their functionality
+              // is handled by JavaScript.
+              if (keyCode == KeyEvent.KEYCODE_BACK || keyCode == KeyEvent.KEYCODE_ESCAPE) {
+                return false
+              }
```
XITE-yusufyildirim, line 294: @cgoldsby Hey Chris, thanks for the great work! I have a question. There are other important `KeyEvent`s which should probably be propagated to the activity like the playback events. So it's important that we don't block those events to not introduce potential issues. You're saying here that returning `false` does propagate events to the activity. The new `onDialogKeyUp` and `onDialogKeyDown` methods you introduced do return false by default already. So, my question is, does that mean every event will be propagated to the activity by default unless people override those methods? That kind of contradicts with the section in your PR description telling there's a breaking change, if I'm not misinterpreting that of course 😄

yusufyildirim, line 294: @cgoldsby Hey Chris, thanks for the great work! I have a question. There are other important KeyEvents which should probably be propagated to the activity like the playback events. So it's important that we don't block those events to not introduce potential issues. You're saying here that returning false does propagate events to the activity. The new onDialogKeyUp and onDialogKeyDown methods you introduced do return false by default already. So, my question is, does that mean every event will be propagated to the activity by default unless people override those methods? That kind of contradicts with the section in your PR description telling there's a breaking change, if I'm not misinterpreting that of course 😄

khoaakt:
I can confirm that this also happens on React Native tvOS 0.75.4-0. Do we have any plans to merge this into 0.75.x?

packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.kt
cgoldsby: Hey @yusufyildirim ,

The implementation is based on a **huge** assumption that I think is correct, but I am not 100% sure:

> In Android, when an `android.app.Dialog` (modal) is shown, it has its own `Window`, which receives the key events rather than routing them through `MainActivity`’s `dispatchKeyEvent`.

For clarity, I will name the two key event handlers:

- **Dialog Window** (present when the Dialog is shown)
- The apps's **MainActivity**

After the changes in this PR, for `ACTION_UP` and `ACTION_DOWN` events while a Dialog is shown:

- Only `KEYCODE_BACK` and `KEYCODE_ESCAPE` will be delegated to the **Dialog Window**.
- All other keys will be delegated to the **MainActivity**.

```diff
- // Allow BACK and ESCAPE keys to propagate to the dialog's activity.
+ // Allow BACK and ESCAPE keys to propagate to the dialog's window.
```

Therefore, the BACK and ESCAPE keys propagate to the **Dialog Window**. If I understood your question correctly, playback key events and all other key events will continue to be delegated to the **MainActivity**.

The potential breaking changes would occur if an app has specific code for `ACTION_UP` in the **MainActivity**.

For example, if you had code to handle a play event in your **MainActivity** that needs to be executed when a Dialog is shown:

```kotlin
// MainActivity.kt

override fun onKeyUp(keyCode: Int, event: KeyEvent?): Boolean {
    return if (keyCode == KeyEvent.KEYCODE_MEDIA_PLAY) {
        // Handle the Play button press
        Log.d("MainActivity", "Play button pressed")

        // Trigger custom logic here
        performPlayAction()

        // Return true to prevent propagation to the system
        true
    } else {
        // Allow other key events to propagate
        super.onKeyUp(keyCode, event)
    }
}
```

In this example, `performPlayAction` would no longer be called when a Dialog is shown. The developer would need to override the `onDialogKeyUp` function to maintain the existing functionality when a dialog is shown:

```kotlin
// MainActivity.kt

override fun onDialogKeyUp(keyCode: Int, event: KeyEvent?): Boolean {
    return if (keyCode == KeyEvent.KEYCODE_MEDIA_PLAY) {
        // Handle the Play button press
        Log.d("MainActivity", "Play button pressed and a dialog is shown")

        // Trigger custom logic here
        performPlayAction()

        // Return true to prevent propagation to the system
        true
    } else {
        // Allow other key events to propagate
        super.onKeyUp(keyCode, event)
    }
}
```

Now `performPlayAction` is called both when the playback button is triggered, regardless of whether a dialog is shown or not.

I hope this example helps. Let me know if I missed something or if you have additional questions!

PR #856, opened 2025-01-13
fix: `disabled` prop - invert condition in Pressability.js event handlers

psysize:
This PR fixes incorrect handling of the `disabled` prop in the `Pressability` class. The `onPressIn` and `onPressOut` handlers now correctly exit early when `disabled` is set to `true`, ensuring that no unintended interactions occur.

I think this mistake slipped in during the #813 refactor.

### DIFF
--- a/packages/react-native/Libraries/Pressability/Pressability.js
+++ b/packages/react-native/Libraries/Pressability/Pressability.js
@@ -440,7 +440,7 @@ export default class Pressability {
   _createEventHandlers(): EventHandlers {
     const tvPressEventHandlers = {
       onPressIn: (evt: any): void => {
-        if (this._config.disabled === false) {
+        if (this._config.disabled === true) {
           return;
         }
 
@@ -461,7 +461,7 @@ export default class Pressability {
         }, delayLongPress + delayPressIn);
       },
       onPressOut: (evt: any): void => {
-        if (this._config.disabled === false) {
+        if (this._config.disabled === true) {
           return;
         }
         this._cancelLongPressDelayTimeout();
--- a/packages/react-native/Libraries/Pressability/__tests__/Pressability-test.js
+++ b/packages/react-native/Libraries/Pressability/__tests__/Pressability-test.js
@@ -241,6 +241,34 @@ describe('Pressability', () => {
     jest.spyOn(HoverState, 'isHoverEnabled');
   });
 
+  describe('disabled', () => {
+    beforeEach(() => {
+      jest.clearAllMocks();
+    });
+
+    it('returns false from onStartShouldSetResponder if disabled is true', () => {
+      const {handlers} = createMockPressability({
+        disabled: true,
+      });
+
+      expect(handlers.onStartShouldSetResponder()).toBe(false);
+    });
+
+    it('prevents press callbacks from being called when disabled is true', () => {
+      const {config, handlers} = createMockPressability({
+        disabled: true,
+      });
+
+      const shouldSetResponder = handlers.onStartShouldSetResponder();
+      expect(shouldSetResponder).toBe(false);
+
+      expect(config.onPressIn).not.toBeCalled();
+      expect(config.onPress).not.toBeCalled();
+      expect(config.onLongPress).not.toBeCalled();
+      expect(config.onPressOut).not.toBeCalled();
+    });
+  });
+
   describe('onBlur', () => {
     it('is called if provided in config', () => {
       const {config, handlers} = createMockPressability();


douglowder:
I agree that this fix is correct, but I'm surprised that you were able to reproduce a real issue on TV. There is native code that should be preventing buttons from being focused at all if they are disabled, so that they can't be selected and `onPress` can't be fired.

I'll test that out and see if there is an additional regression that should be fixed.

RuudBurger:
Don't know if it's related, but we have to do 
```
// RN 0.76 Doesn't like disabled=false for some reason
{...(disabled && { disabled: true })}
```
Sometimes to get buttons working properly. Just doing `disabled={disabled}` or `disabled={false}` would make the button not focusable or clickable.

douglowder:
Apple TV seems to be working as expected if disabled=true. Android TV is not.

Testing this patch, which implements correct behavior if accessible=false on Android TV (note the change in Android native code):

```diff
diff --git a/node_modules/react-native/Libraries/Pressability/Pressability.js b/node_modules/react-native/Libraries/Pressability/Pressability.js
index e70c410..a2b2758 100644
--- a/node_modules/react-native/Libraries/Pressability/Pressability.js
+++ b/node_modules/react-native/Libraries/Pressability/Pressability.js
@@ -440,7 +440,7 @@ export default class Pressability {
   _createEventHandlers(): EventHandlers {
     const tvPressEventHandlers = {
       onPressIn: (evt: any): void => {
-        if (this._config.disabled === false) {
+        if (this._config.disabled === true) {
           return;
         }
 
@@ -461,7 +461,7 @@ export default class Pressability {
         }, delayLongPress + delayPressIn);
       },
       onPressOut: (evt: any): void => {
-        if (this._config.disabled === false) {
+        if (this._config.disabled === true) {
           return;
         }
         this._cancelLongPressDelayTimeout();
@@ -484,12 +484,18 @@ export default class Pressability {
 
     const focusEventHandlers = {
       onBlur: (event: BlurEvent): void => {
+        if (this._config.disabled === true) {
+          return;
+        }
         const {onBlur} = this._config;
         if (onBlur != null) {
           onBlur(event);
         }
       },
       onFocus: (event: FocusEvent): void => {
+        if (this._config.disabled === true) {
+          return;
+        }
         const {onFocus} = this._config;
         if (onFocus != null) {
           onFocus(event);
diff --git a/node_modules/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java b/node_modules/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
index ad5f355..ccc41e9 100644
--- a/node_modules/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
+++ b/node_modules/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
@@ -107,6 +107,7 @@ public class ReactViewManager extends ReactClippingViewManager<ReactViewGroup> {
     if (hasTouchScreen(view.getContext())) {
       view.setFocusableInTouchMode(accessible);
     }
+    view.setFocusable(accessible);
   }
 
   @ReactProp(name = "hasTVPreferredFocus")
```

psysize:
Thanks for getting back to this @douglowder. I'm still wrapping my head around a lot of this, but I tried with your patch above but ran into a few issues:

1. The `setAccessible` method is present in the [core repo](https://github.com/facebook/react-native/blob/559d070e8e4cfc05c7d0798720fddf2180aaa323/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.kt#L80) and our app failed to build because `react-native-svg` was expecting it to be there - I can imagine a few other libs might also expect it to be there - not sure if you where purposeing to remove it or not, but I suppose we could leave it `void` if it turns out the logic is not needed.

2. I tried without the changes to `Pressability.js` but the `this._config.disabled === false` condition still returned from the `pressIn` method early -  maybe we can remove the condition completely if we are solely relying on the `accessible` value, but either way it looks like we will need changes to `Pressability.js`.

3. When modifying your patch to leave `setAccessible` as `void` I ran into issues with `Pressable`s in a vertical list - it seemed to skip some buttons and not others - _I'm still investigating this_.

4. Do we want to be returning early within `onBlur` if a `Pressable` is disabled? I think we would still want to trigger `onBlur` if a focused `Pressable` state changes to `disabled`, otherwise it's locked in a focused state.

_Thanks again, I've been having trouble running the `rn-tester` package and running into some Android caching issues, so building / testing this all is taking a lot of time, but I'll try get back on it._

douglowder:
Thanks @psysize! Given the risks of using my proposed patch, let's just merge your changes to suppress the select action, and we can take some time to test the other changes separately.

douglowder (approved):

PR #910, opened 2025-04-16
[Apple TV] Fix regression in TVEventControl and BackHandler in RN 0.79

douglowder:
Testing with 0.79.1 code today showed that there was a regression on tvOS, where `BackHandler` and `TVEventControl` were not working. As a result, I've made several changes:

- Fix default import of `Platform` in `BackHandler`
- Moved the code in the legacy batched bridge module `RCTTVMenuBridge` to the current native TV event module, for easier future maintenance
- When an event emitter was invoked with no registered listeners, iOS/tvOS was emitting a LogBox warning unnecessarily -- it now just prints an info message to the console


cgoldsby:
The changes look great. I did some tvOS testing with RNTester and definitely saw the back handling working better. Sorry for the delayed response; busy day. 🌞

PR #919, opened 2025-05-06
[Apple TV] Fix for missing select events

### LINKED ISSUE #904: TVEventHandler not sending events for "select" on tvOS
I'm just in the process of upgrading from `react-native-tvos@0.74.3-0` -> `react-native-tvos@0.76.7-0`.

At the moment I have a listener for TV event.type = "select" (when the touchpad area is pressed). It seems that event is no longer firing for me after upgrading. "playPause" and "menu" fire as normal.

Interestingly, on another screen, where there are focusable buttons visible, tapping on those the "select" event fires fine. But on the screen in question, there are no buttons visible when I am listening for the "select" event type.

My set up is like this:
```
  componentDidMount(): void {
      TVEventControl.enableTVMenuKey();
      this._tvEventSubscription = TVEventHandler.addListener(this._handleTVEvent);
      this._backHandlerSubscription = BackHandler.addEventListener(
        'hardwareBackPress',
        this._handleBackPress
      );
    }

  _handleTVEvent = (event: HWEvent): void => {
    // Fires for various vents, but not event.type="select"
    console.log(`TV Event = ${JSON.stringify(event)}`);
  };
```

I browsed through prior discussions and PRs to see if I could work out if I'm doing something wrong or not given the changes that have been made but I couldn't work it out. Related:
https://github.com/react-native-tvos/react-native-tvos/pull/813
https://github.com/react-native-tvos/react-native-tvos/pull/825
https://github.com/react-native-tvos/react-native-tvos/discussions/827

douglowder:
Fixes #904 .

We add a RCTTVRemoteSelectHandler to the root view, and ensure that its gesture recognizer does not run if there is a pressable or touchable present with its own handler.

### DIFF
--- a/packages/react-native/React/Base/RCTRootView.h
+++ b/packages/react-native/React/Base/RCTRootView.h
@@ -14,6 +14,9 @@
 extern NSString * _Nonnull const RCTTVEnableMenuKeyNotification;
 extern NSString * _Nonnull const RCTTVDisableMenuKeyNotification;
 
+#if TARGET_OS_TV
+#import "RCTTVRemoteSelectHandler.h"
+#endif
 
 @protocol RCTRootViewDelegate;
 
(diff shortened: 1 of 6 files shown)


cgoldsby (approved):
🚢 it! 🎉 

The changes look good.

I tested for regressions using a brand new app, the [ExpoRouterTV test app](https://github.com/react-native-tvos/ExpoRouterTV/blob/doug/select-testing/patches/react-native%2B0.76.9-0.patch), and a private app. I did not observe any regressions—only improvements.

I also confirmed the changes compile on iOS (sorry for missing this in previous PRs).

As part of the regression testing, I could no longer reproduce the behavior reported in #904.

PR #921, opened 2025-05-23
[Apple TV] Synchronize focus notifications and direct view events

DrRefactor:
This ensures the same order of focus&blur events for global notifier and direct event listeners. Currently, it's 50/50 for onFocus/onBlur order, while global NSNotification based events are always in deterministic order (onFocus(new) -> onBlur(old)).

Deterministic order makes it easier to handle in custom JS focus mechanisms.

Also, placing onFocus/Blur in coordinatedAnimated ensures visual feedback (parallax animation) is coordinated with onFocus/onBlur listeners making the animation & focus changes more smooth looking.


douglowder:
All right I'll take a look at this and do some testing. This is related to #823 , where I made similar changes to coordinate press animations with onPressIn/onPressOut events. Comments:

- Similarly to #823 , we should probably have onFocus called before animations added, and onBlur called after animations removed.
- Here we also need to handle the case where these animations are disabled.
- Finally, these changes also need to be done for Fabric (`RCTViewComponentView.mm`).

douglowder (approved):

PR #933, opened 2025-06-06
Fix js debugging in multitarget project

DrRefactor:
## Summary:

In multitarget project, CocoaPods generates multiple pod instances for each platform (and/or OS version, swift version etc). See more: https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/target/pod_target.rb

Take a following Podfile as an example:
```
(...)
target 'PlatformPodRepro-iphone' do
  platform :ios, '16.0'
  config = use_native_modules!

  use_react_native!(
    :path => "../node_modules/react-native",
    # An absolute path to your application root.
    :app_path => "#{Pod::Config.instance.installation_root}/.."
  )
end

target 'PlatformPodRepro-appletv' do
  platform :tvos, '16.0'
  config = use_native_modules!

  use_react_native!(
    :path => "../node_modules/react-native",
    # An absolute path to your application root.
    :app_path => "#{Pod::Config.instance.installation_root}/.."
  )

end

post_install do |installer|
  # https://github.com/facebook/react-native/blob/main/packages/react-native/scripts/react_native_pods.rb#L197-L202
  react_native_post_install(
    installer,
    "../node_modules/react-native",
    :mac_catalyst_enabled => false,
    # :ccache_enabled => true
  )
end
```

This causes hermes-engine pod to be split internally into:
- hermes-engine-tvOS
- hermes-engine-iOS

If we define different OS version for one of the targets, pods will become
- hermes-engine-tvOS16.0
- hermes-engine-iOS15.1

This causes following calls to do nothing, as they don't match any pod:
```
self.add_build_settings_to_pod(installer, "GCC_PREPROCESSOR_DEFINITIONS", "HERMES_ENABLE_DEBUGGER=1", "hermes-engine", :debug)
```

Thus, HERMES_ENABLE_DEBUGGER flag is not set and debugging code is disabled in RN iOS code (see InspectorFlags.cpp).

## Changelog:

[IOS] [FIXED] - Fixed new react-native debugger in multitarget iOS projects.

## Test Plan:

Create a fresh project, add separate targets for ios & tvos, use above podfile, build iOS & test.

I'll provide repo with minimal repro shortly.

### DIFF
--- a/packages/react-native/scripts/cocoapods/utils.rb
+++ b/packages/react-native/scripts/cocoapods/utils.rb
@@ -58,7 +58,7 @@ class ReactNativePodsUtils
         # this is needed for Xcode 14, see more details here https://github.com/facebook/react-native/issues/34673
         # we should be able to remove this once CocoaPods catches up to it, see more details here https://github.com/CocoaPods/CocoaPods/issues/11402
         installer.target_installation_results.pod_target_installation_results.each do |pod_name, target_installation_result|
-            if pod_name.to_s == 'React-Core'
+            if target_installation_result.target.root_spec.name == 'React-Core'
                 target_installation_result.resource_bundle_targets.each do |resource_bundle_target|
                     resource_bundle_target.build_configurations.each do |config|
                         config.build_settings['CODE_SIGNING_ALLOWED'] = 'NO'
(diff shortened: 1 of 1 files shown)


douglowder:
@DrRefactor this change seems safe to merge, since it passes React Native core CI tests, so this PR is approved.

Things you should probably be aware of:

- This repo no longer supports Podfiles with multiple targets (as stated in the README). There have been issues where multiple targets would break RN include paths, and it became difficult to maintain support.
- There is work already happening to try and move away from Cocopods towards Swift package manager, although this work is still at a very early stage.

douglowder (approved):

DrRefactor:
Thank you @douglowder

I've missed the one target part. In my project, I have separate targets for iOS and tvOS, and a lot of code that is specific to given platform. Do you suggest merging these targets and replacing "#if TARGET_OS_TV" occurrences with #if available(tvOS, *)"?

Btw, can't wait for SPM to arrive, Pods have been nothing but a headache.

PR #1002, opened 2025-10-03
fix (Android TV): Press events should work when Talkback enabled

### LINKED ISSUE #1001: Pressable onPress does not fire when TalkBack is enabled on AndroidTV
`Pressable` onPress no longer fires when TalkBack is enabled on Android TV. It does work on Android. However, `onAccessibilityAction` fires when used with  `accessibilityActions - activate`.

This is reproducible on 0.82.0-0rc5.

See discussion #1000

douglowder:
## Summary:

The code in `ReactAccessibilityDelegate.kt` was not sending press events correctly when accessibility click actions were detected. The code has been adjusted to send events correctly using the existing PressIn and PressOut events in `ReactViewGroup.kt`.

Fixes #1001 .

This is still in a draft state, as I am not yet able to make long presses work as expected.

## Changelog:

[Android][Fixed] - Press events should work when Talkback enabled

## Test Plan:

- Testing on Android TV emulator and Chromecast
- CI should pass


douglowder:
@zsianko please have a look at this change for the Talkback press events issue. If you are able to build RNTV Android from source, it would be great to get your feedback.

packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactAccessibilityDelegate.kt
```diff
@@ -224,25 +227,16 @@ public open class ReactAccessibilityDelegate( // The View this delegate is attac
...
-      val reactTag = host.id
-      payload.putString("eventType", "select")
-      payload.putInt("eventKeyAction", 1)
-      payload.putInt("tag", reactTag)
-      payload.putInt("target", reactTag)
-      val reactContext = host.context as ReactContext
-      reactContext.emitDeviceEvent("onHWKeyEvent", payload)
+    if (action == AccessibilityNodeInfoCompat.ACTION_CLICK
+      ) {
+      if (hostView is ReactViewGroup) {
```
zsianko, line 232: Shall it be the `host`? I can't see a `hostView` in the scope...Unless I'm missing something

packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactAccessibilityDelegate.kt
```diff
@@ -224,25 +227,16 @@ public open class ReactAccessibilityDelegate( // The View this delegate is attac
     if (action == AccessibilityNodeInfoCompat.ACTION_EXPAND) {
       host.setTag(R.id.accessibility_state_expanded, true)
     }
-    if (action == AccessibilityNodeInfoCompat.ACTION_CLICK) {
```
yusufyildirim: Not sure how these are generating events on the Mobile but not on the TV. I think the same outcome could be achieved by relying on `onAccessibilityEvent` and emitting press/longpress events on the JS layer accordingly.

yusufyildirim (commented):
If you ask me, we shouldn't be firing press/longpress events based on accessibility events in the first place if the upstream RN doesn't do it. People can easily implement it using `onAccessibilityEvent`, if they like to. This way, developers have the ability to distinguish regular press/longpress events vs the accessibility ones. Merging two together takes that flexibility away from them. If you still feel like doing it, I'd say let's not introduce the new native press events and stick to generating them based on `onAccessibilityEvent`.

packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactAccessibilityDelegate.kt
douglowder: Confirmed, it is possible to have press and longPress handlers fire by setting accessibility props correctly, see below.

PR #1054, opened 2026-02-03
fix: tvos request focus on unmounted views

DouweBos:
## Summary:

When we navigate from screen A to screen B we try to move the initial focus to a button a bit further down the page. This is consistently failing ever since adopting the new arch, since `requestTVFocus` is getting called while the component exists but is not part of a view hierarchy yet. When debugging, this came down to `rootView` always being nil, even after the current attempt at skipping a render loop.

So far we have reliably been able to focus these components again with the following changes. This postpones the async re-execution until `didMoveToWindow` is called, under the assumption by that point `containingRootView` will resolve and focussing will succeed.

Of course always happy to get any feedback on this approach.

## Changelog:

[TVOS][FIXED] - Postpone request focus re-execution until view moved to window.


douglowder:
@RuudBurger @DouweBos This fix is included in 0.84.1-0 and 0.83.3-0 releases.

Issue #1061, opened 2026-02-23
TVTextScrollView onFocus/onBlur never fire


slvvn:
### Description

It looks like that after [PR #1050](https://github.com/react-native-tvos/react-native-tvos/pull/1050), **TVTextScrollView**'s `onFocus` and `onBlur` callbacks no longer fire on tvOS. The scroll view still receives focus and scrolls correctly, but there is no way for JS to know when it is focused (e.g. for focus ring style).

## Expected behavior

When a `TVTextScrollView` gains or loses focus, its `onFocus` and `onBlur` props should be called (same as for a regular `View`).

### Steps to reproduce

Run on tvOS, place a TVTextScrollView component with an onFocus prop, and try to focus on it and see that the callback isnt fired.

### React Native Version

0.83

### Affected Platforms

Runtime - iOS

### Output of `npx @react-native-community/cli info`

```text
System:
  OS: macOS 26.3
  CPU: (10) arm64 Apple M1 Pro
  Memory: 169.69 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 22.18.0
    path: /Users/xx/.nvm/versions/node/v22.18.0/bin/node
  Yarn: Not Found
  npm:
    version: 11.7.0
    path: /Users/xx/.nvm/versions/node/v22.18.0/bin/npm
  Watchman:
    version: 2024.12.02.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.16.2
    path: /Users/xx/.rvm/gems/ruby-3.4.1/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.0
      - iOS 26.0
      - macOS 26.0
      - tvOS 26.0
      - visionOS 26.0
      - watchOS 26.0
  Android SDK:
    API Levels:
      - "28"
      - "29"
      - "30"
      - "31"
      - "33"
      - "33"
      - "34"
      - "35"
      - "36"
    Build Tools:
      - 30.0.3
      - 31.0.0
      - 33.0.0
      - 33.0.1
      - 34.0.0
      - 35.0.0
      - 36.0.0
    System Images:
      - android-25 | Google APIs ARM 64 v8a
      - android-28 | Google ARM64-V8a Play ARM 64 v8a
      - android-29 | Polestar 2
      - android-30 | Google APIs ARM 64 v8a
[...]

PR #1062, opened 2026-02-24
fix: restore TVTextScrollView events on tvOS

douglowder:
## Summary:

Fixes regression described in #1061 .

## Test Plan:

Tested that the regression is no longer present in the TVTextScrollView example in RNTester.


cgoldsby (approved):
Thanks for the fix!

I tested this with the rn-tester app using TVTextScrollView, adding console logs to the onFocus and onBlur props. Everything worked as expected.

I made a small commit to align the indentation with the surrounding code.

https://github.com/user-attachments/assets/91e6dffc-3782-4720-b780-4fefff2d42d2

PR #1098, opened 2026-04-07
Support previous/next actions on FireOS

dobladov:
## Summary:

On FireOS `IME_ACTION_NEXT` and `IME_ACTION_PREVIOUS` both fire the `onSubmitEditing` event, but there's no way for the JS side to distinguish between a "submit", "next", or "previous" action.
This makes it impossible to implement proper focus navigation between text inputs on this platform.

This PR adds an `action` field to the `onSubmitEditing` event payload, so developers can handle each case appropriately (e.g., moving focus to the next/previous input vs. submitting a form), in order to be able to handle previous actions on FireOS.

## Changelog:

[ANDROID] [ADDED] - Add action field ('submit' | 'next' | 'previous') to `onSubmitEditing` event for distinguishing IME editor actions in FireOS.

## Test Plan:

Verified that the `onSubmitEditing` event for FireOS now includes the correct action:

- After pressing `previous`
- After submitting the input with `next`
- After pressing back on the remote

Also tested for regressions on Android, no issues were found.

https://github.com/user-attachments/assets/2cdad9b8-ce04-4dd1-9d1b-312e562142b4

Fixes: https://github.com/react-native-tvos/react-native-tvos/issues/1070

PR #1104, opened 2026-04-14
[tvOS] Add action field to onSubmitEditing for tvOS

### LINKED ISSUE #1103: [tvOS] Add action to onSubmitEditing payload
#1098 added an `action` field (`'submit' | 'next' | 'previous'`) to the `onSubmitEditing` event on Android/Fire TV to distinguish IME actions. tvOS doesn't emit it, leaving the field typed as optional and the payload inconsistent across platforms.

**Native (tvOS)**
- Emit `action: "submit"` in Paper — `RCTBaseTextInputView.mm`
- Emit `action: "submit"` in Fabric — `RCTTextInputComponentView.mm`

**JS/TS types**
- Make `action` required in `TextInput.flow.js` and `TextInput.d.ts`
- Drop the `?` on `AndroidTextInputNativeComponent.js`

cgoldsby:
## Summary
Fixes #1103. Companion to #1098.

Aligns the `onSubmitEditing` payload between Android and Apple by including the `action` property on text input submission. The property is non-optional, so cross-platform engineers can rely on it being present. Both Paper and Fabric implementations have been updated. No regression risk since `action` is a new field.

## Changelog
[TVOS] [ADDED] Add `action` field to `onSubmitEditing` event for Apple TV.

## Test Plan
Tested using the `rn-tester` app with the Text Input example. Verified via console logging that `action` is set to `"submit"` on form submission.

### DIFF
--- a/packages/react-native/Libraries/Components/TextInput/AndroidTextInputNativeComponent.js
+++ b/packages/react-native/Libraries/Components/TextInput/AndroidTextInputNativeComponent.js
@@ -429,7 +429,7 @@ export type AndroidTextInputNativeProps = Readonly<{
     Readonly<{
       target: Int32,
       text: string,
-      action?: 'submit' | 'next' | 'previous',
+      action: 'submit' | 'next' | 'previous',
     }>,
   >,
 
--- a/packages/react-native/Libraries/Components/TextInput/TextInput.d.ts
+++ b/packages/react-native/Libraries/Components/TextInput/TextInput.d.ts
@@ -554,6 +554,7 @@ export type TextInputEndEditingEvent =
  */
 export interface TextInputSubmitEditingEventData {
   text: string;
+  action: 'submit' | 'next' | 'previous';
 }
 
 /**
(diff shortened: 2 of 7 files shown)


douglowder (changes requested):
I am not sure that we should add a non-optional payload, as this might be a breaking change from React Native core. Let me do some investigation on this.

douglowder:
@cgoldsby My understanding was that the new property is specifically for Fire TV, and that other Android TV types and Apple TV don't require it. For consistency, we should therefore make the `action` optional.

cgoldsby:
> @cgoldsby My understanding was that the new property is specifically for Fire TV, and that other Android TV types and Apple TV don't require it. For consistency, we should therefore make the `action` optional.

@douglowder, thank you for the review & feedback!

A. Is the suggestion to keep `action` optional in the JS layer, and keep the tvOS changes?
B. Or, drop the `action` from being sent natively on tvOS, keeping `action` optional, therefore closing the PR?

Either option is 💯% a-okay with me.

A little more details from my side:
- The most important thing for me is having the `action` property set for Fire TV and Android TV. (#1098)
- This PR is a _nice-to-have_ (from my perspective). And, I personally find having the action set all TV platforms more consistent.

douglowder:
> A. Is the suggestion to keep `action` optional in the JS layer, and keep the tvOS changes?
> B. Or, drop the `action` from being sent natively on tvOS, keeping `action` optional, therefore closing the PR?

@cgoldsby the action should be optional on both platforms, to keep compatibility with RN core. Since tvOS isn't really using the action, I think option B is correct.

PR #1118, opened 2026-06-05
fix(tvos): continuous scroll animation on snap

koenoe:
**Problem**
- On a tvOS `snapToAlignment="item"` list, the scroll animation stutters when holding the d-pad instead of running as one smooth, continuous animation.

**Cause**
- On each focus change the focus engine already scrolls to the snap offset — `scrollViewWillEndDragging` redirects its scroll to `preferredContentOffset` (added in #1068). The snap *also* ran an imperative `setContentOffset:` to the same offset: a second, duplicate scroll that competes with the engine's own scroll/animation, and that interference is the stutter.

**Fix**
- Remove the imperative `setContentOffset:`; setting `preferredContentOffset` and letting the engine's scroll position the item is enough.
- That's what the delegate is for — per Apple: *"Focus movements still cause this delegate method to be called as you would expect, and the targetContentOffset parameter can be used to specify an alternative final content offset."* ([Apple Frameworks Engineer](https://developer.apple.com/forums/thread/22103?answerId=73300022#73300022) · [`scrollViewWillEndDragging` docs](https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/scrollviewwillenddragging(_:withvelocity:targetcontentoffset:)))
- `scrollAnimationEnabled={false}` keeps the instant path (the engine's reveal is suppressed there); tvOS-only, Android unchanged.

**AI usage**

Code-path and PR-history investigation and the change itself were produced with AI assistance (Claude Code), then human-reviewed and validated on a physical tvOS device.

### DIFF
--- a/packages/react-native/React/Fabric/Mounting/ComponentViews/ScrollView/RCTScrollViewComponentView.mm
+++ b/packages/react-native/React/Fabric/Mounting/ComponentViews/ScrollView/RCTScrollViewComponentView.mm
@@ -1283,8 +1283,20 @@ static inline UIViewAnimationOptions animationOptionsWithCurve(UIViewAnimationCu
     CGPoint targetContentOffset = isHorizontalSnap
         ? CGPointMake(targetOffset, scrollView.contentOffset.y)
         : CGPointMake(scrollView.contentOffset.x, targetOffset);
+    // Where the focused item should snap to. We don't scroll to it here: this view's
+    // scrollViewWillEndDragging:withVelocity:targetContentOffset: override sets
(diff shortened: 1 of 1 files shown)


koenoe:
@douglowder apologies for possible confusion by closing and reopening, but I'm happy with the current state. Is there anything left on my side that I can do to get this reviewed and merged? Thank you!

douglowder:
Sorry for the delay, we will take a look.

douglowder (commented):
This code looks good. It would be very helpful if you could add to the PR description, and explain how to test this with the existing snap scroll examples in RNTester.

RuudBurger (approved):

PR #792, opened 2024-09-28
feat: make TVFocusGuideView work with Talkback

mateusz1913:
Talkback handles focus on its own, so neither "autoFocus", nor "destinations" concepts of TVFocusGuideView were not working on AndroidTV with Talkback enabled.

To implement, the custom ReactAccessibilityDelegate needs to be implemented and it needs to use focus guide utilities from ReactViewGroup, to correctly forward the focus from focus guide container to its child focusable views.

This is only needed for Talkback and implemented logic will not be applied to other accessibility services, e.g. Amazon's VoiceView (which handles TVFocusGuideView concepts out of the box)

When developing this PR, I was using Google TV Chromecast with Android 12

https://github.com/user-attachments/assets/5cf648ef-ed49-4073-84b2-6875630b4ba8

If this PR would be accepted, I'll follow up with another one for branch v0.76

PR #1132, opened 2026-06-24
[Android TV] fix talkback focus guide handling after Kotlin conversion

mateusz1913:
After `ReactViewGroup` conversion to Kotlin, the logic to intercept a11y delegate events became wrongly executed only when there is non-null `Bundle` argument - as a result, the `destinations` or last focused element stopped receiving the events

The previous logic implemented in java didn't care about the nullability of `Bundle` argument, as it was always just forwarded to `performAccessibilityAction` on the target views, so to fix and maintain correct behavior, the `handleInterceptEvents` signature now expects optional `Bundle` args argument

## Summary:

We're upgrading our client's React Native TV app from 0.79 to 0.85 and it turned out that the focus guide talkback handling (that I contributed 2 years ago in https://github.com/react-native-tvos/react-native-tvos/pull/792) stopped working after the `ReactViewGroup` was converted to Kotlin file (initial focus guide handling implementation was in java)

## Changelog:

[ANDROID] [FIXED] - talkback focus guide handling after Kotlin conversion

## Test Plan:

Run RNTester focus guide example with Talkback before and after the change. Before the change the focus will be stuck on the focus guide container, after the change it will correctly redirect the focus to `destinations` or last focused element

### DIFF
--- a/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.kt
+++ b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.kt
@@ -1174,7 +1174,7 @@ public open class ReactViewGroup public constructor(context: Context?) :
           return super.performAccessibilityAction(host, action, args)
         }
         if (action == AccessibilityNodeInfo.ACTION_FOCUS) {
-          if (args?.let { host.interceptAccessibilityEvents(action, it) } == true) {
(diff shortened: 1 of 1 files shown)


douglowder (approved):

douglowder (changes requested):
Please rebase this to the main branch, and we can get this landed.

mateusz1913:
Done, thanks for the review!

douglowder (approved):

PR #1141, opened 2026-06-28
Add onFocusEnter / onFocusLeave to TVFocusGuideView

RuudBurger:
TV `onFocus`/`onBlur` bubble, so a container fires them for every descendant focus change, including focus moving between two children inside it. There was no clean way to know "focus entered this subtree" / "focus left this subtree" without counting events yourself. These two props on `TVFocusGuideView` do exactly that.

`onFocusEnter` fires once when focus crosses into the guide's subtree, `onFocusLeave` once when it fully leaves. Moving focus between descendants doesn't re-fire either. Same boundary-crossing idea as `onMouseEnter`/`onMouseLeave` vs `onMouseOver`/`onMouseOut`. New prop names because `onFocus`/`onBlur`/`onFocusCapture`/`onBlurCapture` are all taken.

Scoped to `TVFocusGuideView` on purpose: it's the focus container, so that's where this belongs, and it keeps plain `View` untouched (no extra hooks/handlers on the most-instantiated component in the app).

## How it works
- Pure JS, no native changes. Works on tvOS and Android TV since both already bubble these events.
- A small `useFocusEnterLeave` hook hooks the **capture** phase (`onFocusCapture`/`onBlurCapture`) on the guide's inner View, not bubble, so a descendant calling `stopPropagation` can't desync it. Any `onFocus`/`onBlur`/`onFocusCapture`/`onBlurCapture` you pass through still work.
- Enter/leave is tracked with a boolean plus a deferred leave: a child to child move fires blur then focus back to back, and the incoming focus cancels the pending leave, so nothing fires. State resets on unmount.

## What's here
- `useFocusEnterLeave` hook + wiring in `TVFocusGuideView`
- `onFocusEnter`/`onFocusLeave` on `TVFocusGuideView`'s Flow and TS types
- rn-tester example (Focus Enter / Leave) showing bubbled counts spiking on every move vs enter/leave firing once
- Unit tests for the hook (enter-once, leave-once, child to child coalescing, re-fire after a real leave, user handlers chained, unmount reset)

## Verified
jest (5/5), flow, eslint, and tsc all clean (no net-new errors; `View` is back to untouched).

### DIFF
--- a/packages/react-native/Libraries/Components/TV/TVFocusGuideView.js
+++ b/packages/react-native/Libraries/Components/TV/TVFocusGuideView.js
@@ -8,6 +8,7 @@
  * @format
  */
 
+import type {BlurEvent, FocusEvent} from '../../Types/CoreEventTypes';
 import type {ViewProps} from '../View/ViewPropTypes';
 import type {ComponentOrHandleType} from './tagForComponentOrHandle';
 
@@ -16,6 +17,7 @@ import useMergeRefs from '../../Utilities/useMergeRefs';
 import View from '../View/View';
 import {Commands} from '../View/ViewNativeComponent';
 import tagForComponentOrHandle from './tagForComponentOrHandle';
+import useFocusEnterLeave from './useFocusEnterLeave';
 
 const StyleSheet = require('../../StyleSheet/StyleSheet').default;
 const React = require('react');
@@ -51,6 +53,20 @@ type TVFocusGuideViewProps = $ReadOnly<{
    * When set to false, this view and all its subviews will be NOT focusable.
    */
   focusable?: boolean | void,
+
+  /**
+   * Fires once when focus enters this focus guide's subtree from outside.
+   * Unlike `onFocus`, which bubbles and fires for every descendant, this does
+   * not fire again when focus moves between descendants. Boundary-crossing
+   * only, same idea as `onMouseEnter` vs `onMouseOver`.
+   */
+  onFocusEnter?: ?(event: FocusEvent) => void,
+
+  /**
+   * Fires once when focus leaves this focus guide's subtree entirely. Does
+   * not fire when focus moves between descendants. Pairs with `onFocusEnter`.
+   */
+  onFocusLeave?: ?(event: BlurEvent) => void,
 }>;
 
 export type TVFocusGuideViewImperativeMethods = $ReadOnly<{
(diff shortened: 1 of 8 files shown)


packages/react-native/Libraries/Components/TV/useFocusEnterLeave.js
```diff
@@ -0,0 +1,96 @@
...
+  onFocusEnter: ?(event: FocusEvent) => void,
+  onFocusLeave: ?(event: BlurEvent) => void,
+  onFocusCapture: ?(event: FocusEvent) => void,
+  onBlurCapture: ?(event: BlurEvent) => void,
+): FocusEnterLeaveHandlers {
+  const stateRef = React.useRef<FocusEnterLeaveState>({
+    entered: false,
+    pendingLeave: null,
+  });
+
```
douglowder, line 45: If both `onFocusEnter` and `onFocusLeave` are null here, we could just return
```
{
  handleFocusCapture: onFocusCapture,
  handleBlurCapture: onBlurCapture,
}
```

packages/react-native/Libraries/Components/TV/TVFocusGuideView.js
```diff
@@ -103,6 +123,16 @@ function TVFocusGuideView({
...
 
+  // Collapse the bubbled per-descendant focus/blur into a single enter when
+  // focus crosses into the guide's subtree and a single leave when it exits.
+  const {handleFocusCapture, handleBlurCapture} = useFocusEnterLeave(
+    onFocusEnter,
+    onFocusLeave,
+    onFocusCapture,
+    onBlurCapture,
+  );
+  const trackFocusEnterLeave = onFocusEnter != null || onFocusLeave != null;
```
douglowder: It seems to me that the `useFocusEnterLeave` hook could just return `onFocusCapture` and `onFocusLeave` when both `onFocusEnter` and `onFocusLeave` are null, correct? Then we could always return the handlers below. See my comment in `useFocusEnterLeave`

douglowder (commented):
I added some questions inline

douglowder:
@RuudBurger I tested this locally, looks good. I created a branch and added a couple of commits, to shortcut the effect if there are no handlers defined, and to fix the metadata in the RNTester example. https://github.com/react-native-tvos/react-native-tvos/compare/ruud/view-focus-enter-leave...doug/view-focus-enter-leave

douglowder (approved):

PR #1144, opened 2026-07-29
CI: fix monorepo issues (package renaming issues, unit tests)

douglowder:
Several commits to fix issues with package renaming of `react-native` and `virtualized-lists`.

- Allow unit tests to run without manually renaming packages
- Ensure Typescript generation and tests work in all cases
- Fix release build breakage caused by the restoration of `react-native/set-env` usage in https://github.com/react/react-native/pull/57560

## Test Plan

Unit tests and build workflows should succeed.


tools/rntv-workflows/scripts/run-tv-unit-tests.sh
```diff
@@ -4,7 +4,7 @@
 # Fail if anything errors
 set -eox pipefail
 
-./tools/rntv-workflows/rename-virtualized-lists.sh
+#./tools/rntv-workflows/rename-virtualized-lists.sh
```
cgoldsby: Since the script was deleted in this PR, we can safely remove this line.

```suggestion
```

cgoldsby (approved):

PR #1148, opened 2026-08-11
feat(android): configurable scroll animation duration and easing

koenoe:
**Problem**
- TV apps design their browse experience around the scroll feel, but on Android TV the snap scroll animation is fixed: the OverScroller default duration (~250ms) with the default `AccelerateDecelerateInterpolator`. The only lever is `scrollAnimationEnabled`. Matching a design language's scroll feel is impossible from JS.

**Cause**
- On tvOS the focus engine performs the snap scroll, so timing and easing are the platform's own (#1068, #1118). On Android the snap scroll is driven by the fork itself — `startFlingAnimator` runs an `ObjectAnimator` with a hardcoded duration and never sets an interpolator — so the platform never gets a say, and neither does the app.

**Fix**
- Add two ScrollView props, Android TV only, additive: `scrollAnimationDuration` (ms) and `scrollAnimationEasing` (CSS easing keyword or cubic-bezier `[x1, y1, x2, y2]`). Keywords resolve to their CSS bezier values in JS; native always receives four numbers and builds a `PathInterpolator`. Unset props keep today's behavior exactly.
- The interpolator is re-applied on every `startFlingAnimator` call (the animator instance is reused), and both props reset in `initView()` so recycled views don't inherit another view's timing.
- The RNTester TV scroll-animation example now shows rows with contrasting settings side by side. tvOS unchanged: the focus engine keeps owning the animation there.

**Changelog**

[ANDROID] [ADDED] - `scrollAnimationDuration` and `scrollAnimationEasing` props on ScrollView for Android TV

**AI usage**

Investigation, the change, and the RNTester example were produced with AI assistance (Claude Code), then human-reviewed and tested.

https://github.com/user-attachments/assets/ee88c899-4d01-4463-be1d-0795953b8ad8

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,7 +316,7 @@ class Game2048 extends React.Component {
   </ScrollView>
   ```
 
-  - _ScrollView animation timing (Android TV)_: The `scrollAnimationDuration` and `scrollAnimationEasing` props tune the animated scroll that runs when focus moves to another item. Both are Android TV only; on tvOS the focus engine owns this animation. When either prop is unset, the platform defaults are used, so behavior is unchanged.
+  - _ScrollView animation timing (Android TV)_: The `scrollAnimationDuration` and `scrollAnimationEasing` props tune the animated scroll that runs when focus moves to another item. They apply to the snap scroll (`snapToAlignment="item"` with per-item `scrollSnapAlign`/`scrollSnapOffset` markers) and to animated `scrollTo` commands; other focus scrolls are instant. Both are Android TV only; on tvOS the focus engine owns this animation. When either prop is unset, the platform defaults are used, so behavior is unchanged.
 
   | Prop (ScrollView) | Value | Description |
   |---|---|---|
--- a/README.md
+++ b/README.md
@@ -316,6 +316,20 @@ class Game2048 extends React.Component {
   </ScrollView>
   ```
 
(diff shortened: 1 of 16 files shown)


douglowder:
@koenoe I built RNTester with your changes, and increase the animation duration to 2000 ms for the scroll animation examples, and do not see the expected slow transitions on my emulator.

koenoe:
> @koenoe I built RNTester with your changes, and increase the animation duration to 2000 ms for the scroll animation examples, and do not see the expected slow transitions on my emulator.

@douglowder 
Fixed the example in this commit: https://github.com/react-native-tvos/react-native-tvos/pull/1148/changes/8f7383cca9130f929136dc4bdefedca5e2f72082

The transition you saw is Android's default scroll animation when an item that receives focus and is outside the viewport, scrolls into the viewport. Scroll snapping can't use that animation, because unlike TVOS (where we can set the `targetContentOffset` on the focus engine) there's nothing in Android to hook into. This is why we have a custom animator instead (introduced in #1068).

The examples I created weren't opted into snapping, so they weren't using the custom snap animation. Now they are and I've updated the video to reflect this.

douglowder (approved):
Looks good! Waiting on @yusufyildirim for review of Android native changes

RuudBurger (approved):

PR #1153, opened 2026-09-02
fix(android): lay unconstrained text out at hintWidth again (0.77)

jash90:
## Summary

#899 ("Add Android API 22 support to version 0.77", shipped in 0.77.2-0 … 0.77.3-1) replaced `hintWidth` with `(int) width` in the non-boring branch of `ReactTextShadowNode.measureSpannedText`. When Yoga measures the node with an undefined width — e.g. children of a horizontal `ScrollView`/`FlatList`, or any content-sized text without a width constraint — `width` is `NaN`, `(int) NaN` is `0` and the `StaticLayout` wraps one glyph per line. With `numberOfLines={1}` only the ellipsis survives.

Latin text is unaffected because `BoringLayout.isBoring` succeeds and takes the other branch; Hebrew/Arabic (and any text `BoringLayout` rejects) breaks. Upstream 0.77.3, 0.77.0-0/0.77.1-0 of this fork and 0.78.0-0+ all use `hintWidth`; only the 0.77.2/0.77.3 line has the regression. Legacy architecture only (Fabric goes through `TextLayoutManager`).

## Repro

Old architecture, Android TV, `I18nManager.forceRTL(true)`:

```jsx
<FlatList horizontal data={[{ id: '1' }]} renderItem={() => <Text>לצפייה</Text>} />
```

0.77.3-1 renders one letter per line in a narrow tall box; with this change (and on 0.77.1-0 / 0.78.0-0) it renders on one line.

## Changelog

[ANDROID] [FIXED] - Fix non-boring text wrapping one character per line when measured without a width constraint (legacy architecture)

## Test Plan

Verified on an Android TV app (react-native-tvos 0.77.3-1, legacy arch, forced RTL) by routing the same measurement through the upstream measure function with a finite width: the detail-screen buttons, side-menu labels and settings tabs render on one line again; Latin text unchanged.


douglowder:
Android 22 support is now in the 0.83 branch (`tvos-v0.83.0`). If this issue is still happening there, in the new architecture, you can rebase this PR against that branch, and we will take a look. We no longer support the old architecture.

douglowder (changes requested):
Android 22 support is now in the 0.83 branch (`tvos-v0.83.0`). If this issue is still happening there, in the new architecture, you can rebase this PR against that branch, and we will take a look. We no longer support the old architecture.

